PR #15, opened 2020-01-21
Support opaque types that are not structs

### LINKED ISSUE #13: Support opaque C++ types that are not structs
For example we might have:

```cpp
// c++

using Obj = void*;
```

```rust
// rust

#[cxx::bridge]
mod ffi {
    extern "C" {
        type Obj;
        fn f(obj: &Obj);
    }
}
```

Currently this fails to compile because we forward declare opaque C types assuming they are structs.

```console
target/debug/build/cxxtest/out/src/main.rs.cc:5:8: error: using typedef-name ‘using Obj = void*’ after ‘struct’
    5 | struct Obj;
      |        ^~~
```

dtolnay:
Fixes #13.

This way we can have:

```cpp
// c++

using Obj = void*;
```

```rust
// rust

#[cxx::bridge]
mod ffi {
    extern "C" {
        include!("...");

        type Obj;
        fn f(obj: &Obj);
    }
}
```

We no longer assume anything about opaque C++ types except that something with the right name has been declared by one of the header(s).

cc @not-a-seagull

### DIFF
--- a/gen/write.rs
+++ b/gen/write.rs
@@ -40,7 +40,8 @@ pub(super) fn gen(namespace: Vec<String>, apis: &[Api], types: &Types, header: b
     for api in apis {
         match api {
             Api::Struct(strct) => write_struct_decl(out, &strct.ident),
-            Api::CxxType(ety) | Api::RustType(ety) => write_struct_decl(out, &ety.ident),
+            Api::CxxType(ety) => write_struct_using(out, &ety.ident),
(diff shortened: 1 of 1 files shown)


not-a-seagull:
Well you're fast... :smile: 

I think this would solve the problem I've been having.

dtolnay:
I published 0.1.2 containing this fix.

PR #32, opened 2020-02-03
(WIP) Bazel CI

sayrer:
This might get a little noisy as I make sure this stuff works on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ addons:
       - bazel
   homebrew:
     taps: bazelbuild/tap
-    casks: bazel
+    packages: bazelbuild/tap/bazel
 
 install:
   - cargo install cargo-raze
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,17 @@ os:
   - linux
   - macos
 
+install:
+  - cargo install cargo-raze
+  - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt install curl && \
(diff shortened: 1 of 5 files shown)


sayrer:
OK, this build installs Bazel, cargo-raze and generates build files, but doesn't actually do anything with Bazel yet. Will update.

.travis.yml
```diff
@@ -6,6 +6,21 @@ os:
...
+      - sourceline: 'deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8'
+        key_url: 'https://bazel.build/bazel-release.pub.gpg'
+    packages:
+      - bazel
+  homebrew:
+    taps: bazelbuild/tap
+    packages: bazelbuild/tap/bazel
+
+install:
+  - cargo install cargo-raze
```
dtolnay: According to https://travis-ci.com/dtolnay/cxx/jobs/283619219 this step added more than 10 minutes to the build. That's crazy! Is there some other way without building this from source? If not, could you coordinate with the cargo-raze maintainers to publish official binaries for CI?

sayrer: Yeah, I noticed that too. I view this more as a problem with Cargo itself, but not one this project should take on.

I'll work on this patch until it's ready and get to your legitimate point at the end.

luser: It's straightforward to ask Travis to [cache the cargo directory](https://docs.travis-ci.com/user/languages/rust/#dependency-management) which would make this fast in subsequent builds at least.

dtolnay: I have had enough bad experiences with Travis cache that I have no interest in using it. It is slow and flaky, and with the amount of junk Cargo puts in there over time it ends up taking longer to fetch the cache than what it saves you.

sayrer: Yes, I agree with dtolnay's assessment here. I think it would be better to get the cargo-raze repo to publish binaries.

PR #59, opened 2020-03-04
Add a second Bazel CI based on cargo-raze

sayrer:
I wanted to take another look at this, because I don't think the Bazel setup that's on master will work well for external projects, because of the way the vendoring works. I think "cargo raze" installation should be faster now. So, I'll just open this PR to take a look on Travis.


dtolnay:
Sure thing. I am going to mute this thread while you iterate &mdash; please close and open another PR when you need me to take a look.

sayrer:
Looks like this is still slow, because cargo-raze has added new code that should compile faster, but hasn't yet removed the old code: https://github.com/google/cargo-raze/issues/132

sayrer:
This is down to ~4 minutes if cargo-raze is installed in debug mode (which should be fine, since the command takes 2.5 seconds to run when compiled that way). I'll try to get cargo-raze to add a feature that eliminates the cargo dependency next.

sayrer:
It looks like this will be about as fast the other builds if I can get the cargo-raze changes accepted.  Right now, the Bazel CI build is only a little slower than the macOS build, and it's running the cargo-raze stuff as well as the existing Bazel code.

sayrer:
A patch similar to mine eventually got landed in `cargo-raze`, so it should build much faster now.

I am still making a separate `cc_library` from the `rust_library` in my own code, so that both are compiled with the toolchains specified in the Bazel workspace. The only thing I had to add was 
a linkstatic=True to the extra rule.  I'll just need to decide whether to add little more recombination in the appended Bazel code.

```
#!/usr/bin/env bash

cargo generate-lockfile && \
cargo vendor --versioned-dirs --locked && \
cargo raze && \
cat >> $(find vendor/cxx-* -name "BUILD.bazel") <<- EOM

cc_library(
 name="cxx_cc_library",
 hdrs=glob(["include/**/*.h"]),
 srcs=glob(["src/**/*.cc"]),
 linkstatic=True,
)

EOM
```

dtolnay (approved):
Thanks for getting this working! I think I would prefer not to have this checked into the repo, but it will be here for anyone searching for cargo-raze guidance (https://github.com/dtolnay/cxx/search?q=cargo-raze&type=issues).

PR #67, opened 2020-03-13
C++ std::vector<T> and Rust std::vec::Vec<T> support

myronahn:
Implements #33.

Add basic std::vector<T> and std::vec::Vec<T> support across FFI boundary.

PR #96, opened 2020-04-10
Fix ssize_t on Windows

AlesTsurko:
Rust's `isize` converted into `ssize_t`, but it's not defined on Windows. That leads to compilation errors. There's `SSIZE_T` though.

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -8,6 +8,10 @@
 #include <type_traits>
 #include <utility>
 
+#ifdef _WIN32
+#define ssize_t SSIZE_T
+#endif
+
 namespace rust {
 inline namespace cxxbridge02 {
 


myronahn:
I have a similar fix in my pull request.  It does seem to fix the problem although it could be that the header just happened to be included in all the right places.

https://github.com/dtolnay/cxx/pull/67/commits/15b5b10763219ba6386abae073a11dd6b06a8ea1

dtolnay:
Ah I guess this relies on at least one of the `include!`'d files to contain `#include "rust/cxx.h"`, which wouldn't always be the case.

Is there a meaningful difference between using `_WIN32` vs `defined(_MSC_VER)`?

Alternatively we could do this based on `env::var_os("CARGO_CFG_WINDOWS").is_some()` for Cargo builds since Cargo guarantees that is passed for all Windows targets, and use something like a `--windows` flag for the cxxbridge cli command for the case of non-Cargo builds.

dtolnay:
I think I prefer @myronahn's use of a type alias instead of a macro for this. But I would prefer not to have our header define `ssize_t` in the root namespace. Could we define it as `rust::isize` instead, using `ssize_t` as the type in the non-Windows case?

dtolnay:
I've published cxx 0.2.7 containing the fix from #97.

PR #97, opened 2020-04-10
Define rust::isize with Windows support

dtolnay:
Hybrid of https://github.com/dtolnay/cxx/commit/15b5b10763219ba6386abae073a11dd6b06a8ea1 and https://github.com/dtolnay/cxx/commit/ef1bf1822c243a38832602b75d21d3da1fd77804 and https://github.com/dtolnay/cxx/pull/96#issuecomment-612150188.

cc @myronahn @AlesTsurko
closes #96

### DIFF
--- a/gen/include.rs
+++ b/gen/include.rs
@@ -38,6 +38,7 @@ pub struct Includes {
     pub string: bool,
     pub type_traits: bool,
     pub utility: bool,
+    pub base_tsd: bool,
 }
 
 impl Includes {
@@ -87,6 +88,11 @@ impl Display for Includes {
         }
         if self.utility {
             writeln!(f, "#include <utility>")?;
+        }
+        if self.base_tsd {
+            writeln!(f, "#if defined(_WIN32)")?;
+            writeln!(f, "#include <BaseTsd.h>")?;
+            writeln!(f, "#endif")?;
         }
         if *self != Self::default() {
             writeln!(f)?;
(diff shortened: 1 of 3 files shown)


dtolnay:
Following up on `_WIN32` vs `_MSC_VER` from https://github.com/dtolnay/cxx/pull/96#issuecomment-612146914: it sounds like `_WIN32` is defined when building *for* windows with any compiler, while `_MSV_VER` is defined when building specifically with the Microsoft compiler. Unclear which is more appropriate in this case but I would think `_WIN32` because building the same code with a different compiler on windows isn't going to cause a different set of headers to be included (is it?). @myronahn @AlesTsurko I would appreciate any insight on this.

AlesTsurko:
Really can't say something reasonable about it. I'm not an expert in the Windows development.

Anyway, if this is true:
> _WIN32 is defined when building for windows with any compiler, while _MSV_VER is defined when building specifically with the Microsoft compiler.

the `_WIN32` is the correct choice.

I used the Visual Studio's/Microsoft compiler to check #96

myronahn:
I'm not an expert either but `_WIN32` sounds like the better choice.

myronahn:
I believe `BaseTsd.h` is in the Windows SDK and so theoretically independent of the Microsoft Compiler.  `_WIN32` is probably more appropriate in this case.

PR #101, opened 2020-04-11
Accept RUST_CXX_NO_EXCEPTIONS to disable throwing

dtolnay:
Partially addresses #92 (it remains to figure out how to nicely expose constructing a rust::String from a std::string when exceptions are off).

### DIFF
--- a/src/cxx.cc
+++ b/src/cxx.cc
@@ -1,8 +1,19 @@
 #include "../include/cxx.h"
+#include <cstdlib>
 #include <cstring>
 #include <iostream>
 #include <memory>
 #include <stdexcept>
+
+template <typename Exception>
+[[noreturn]] static void panic(const char *msg) {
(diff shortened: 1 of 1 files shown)


dtolnay:
Tested by adding `.flag("-fno-exceptions")` in [build.rs](https://github.com/dtolnay/cxx/blob/0.2.7/build.rs), confirming that it fails to compile, then adding `.define("RUST_CXX_NO_EXCEPTIONS", None)` and confirming that it then compiles.

slurps-mad-rips:
@dtolnay might be a better choice to use [`std::terminate`][1] over `abort`, as someone can then replace the `terminate` handler, and possibly get a stack trace via libunwind or some other means before abort is immediately called. 

[1]: https://en.cppreference.com/w/cpp/error/terminate

dtolnay:
Fixed in #112 to use std::terminate. Good call!

Issue #51, opened 2020-03-02
Support associated methods


dtolnay:
Currently only freestanding functions are supported by this crate, not methods.

I like the approach of using `self` in the foreign function parameter list to signify a method receiver:

```rust
#[cxx::bridge]
mod ffi {
    extern "C" {
        type C;
        fn f(self: &C);
    }

    extern "Rust" {
        type R;
        fn g(self: &R);
    }
}

struct R {...}

impl R {
    fn g(&self) {...}
}
```

This would expose the methods `r.g()` to C++ and `c.f()` to Rust.

The above syntax would require rustc 1.43 or newer for https://github.com/rust-lang/rust/pull/68728.

myronahn:
Would it make sense to use the `&self` shorthand instead of `self: &C`?

dtolnay:
That would be fine if there is only one extern type in the block.

dtolnay:
Implemented in #121.

I filed #124 to follow up on `&self` and `&mut self` shorthand inside of extern blocks that contain exactly one extern type.

PR #121, opened 2020-04-16
Support calling C++ and Rust methods

jgalenson:
This adds support for calling C++ and Rust methods with the syntax suggested in #51:

fn get_name(self: &ThingC) -> &CxxString;
fn print(self: &ThingR);

It requires Rust 1.43 to use this syntax.

### DIFF
--- a/gen/write.rs
+++ b/gen/write.rs
@@ -326,6 +326,8 @@ fn write_struct_with_methods(out: &mut OutFile, ety: &ExternType, methods: &Vec<
         writeln!(out, "//{}", line);
     }
     writeln!(out, "struct {} final {{", ety.ident);
+    writeln!(out, "  {}() = delete;", ety.ident);
+    writeln!(out, "  {}(const {}&) = delete;", ety.ident, ety.ident);
     for method in methods {
         write!(out, "  ");
         let sig = &method.sig;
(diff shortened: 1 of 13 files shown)


dtolnay:
Thanks, this is great! I'll get a chance to review the implementation tomorrow.

If someone needs method support on 1.42 and isn't satisfied waiting for the 1.43 release, it would be possible for us to accept `this` as a synonym for `self`. But I would lean toward sticking with just `self` until someone asks for it.

myronahn:
This is great, I can really use this.

gen/write.rs
```diff
@@ -300,6 +321,21 @@ fn write_struct_using(out: &mut OutFile, ident: &Ident) {
...
+    }
+    writeln!(out, "struct {} final {{", ety.ident);
+    for method in methods {
+        write!(out, "  ");
+        let sig = &method.sig;
+        let local_name = method.ident.to_string();
+        write_rust_function_shim_decl(out, &local_name, sig, None, false);
+        writeln!(out, ";");
+    }
+    writeln!(out, "}};");
```
dtolnay, line 336: We need a private default constructor in here. Otherwise this allows C++ to obtain an opaque Rust type by value and call methods on it, which is unsound.

Specifically the R2 struct from tests/ffi/lib.rs currently turns into:

```cpp
struct R2 final {
  size_t get() noexcept;
  size_t set(size_t n) noexcept;
};
```

and this allows C++ to write `R2{}.get()` which becomes UB.

jgalenson: Thanks for catching that!

jgalenson:
I marked it as a draft because I thought we wanted to wait until Rust 1.43 was released, but it looks like you dealt with that as a follow-up.  Thanks!

PR #131, opened 2020-04-22
Allow &self without a type when the block only has one type

### LINKED ISSUE #124: Support &self shorthand if there is only one extern type in the block
As currently implemented, Rust associated methods and C++ member functions are currently specified with `self: &T` syntax.

```rust
extern "C" {
    type C;
    fn f(self: &C);
}
```

Inside of extern blocks that contain exactly one extern type, we should allow the usual `&self` and `&mut self` shorthand as well.

```rust
extern "C" {
    type C;
    fn f(&self);
}
```

jgalenson:
Fixes #124.

### DIFF
--- a/syntax/parse.rs
+++ b/syntax/parse.rs
@@ -91,16 +91,28 @@ fn parse_foreign_mod(foreign_mod: ItemForeignMod) -> Result<Vec<Api>> {
         Lang::Rust => Api::RustFunction,
     };
 
-    let mut items = Vec::new();
+    let mut types = Vec::new();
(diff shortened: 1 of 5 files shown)


dtolnay:
Published to crates.io in 0.2.10.

dtolnay:
@jgalenson, if you care about these commits being attributed to your account in https://github.com/dtolnay/cxx/graphs/contributors (up to you), you would need to add your work email to https://github.com/settings/emails.

jgalenson:
Thanks for the tip; that can't hurt!

PR #157, opened 2020-04-27
Add all std::iterator_traits required types to const_iterator of rust::Vec<T>

myronahn:
https://en.cppreference.com/w/cpp/iterator/iterator_traits

`std::iterator_traits` requires the following 5 types:

- `difference_type` - a signed integer type that can be used to identify distance between iterators
- `value_type` - the type of the values that can be obtained by dereferencing the iterator. This type is void for output iterators.
- `pointer` - defines a pointer to the type iterated over (value_type)
- `reference` - defines a reference to the type iterated over (value_type)
- `iterator_category` - the category of the iterator. Must be one of iterator category tags.

The current `const_iterator` for `rust::Vec<T>` only defined `value_type` and `reference` which caused an error when using `std::copy` on gcc 7.5.0 on Ubuntu but seemed to work fine on MacOS.

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -235,9 +235,13 @@ public:
 
   class const_iterator {
   public:
+    using difference_type = ptrdiff_t;
     using value_type = typename std::add_const<T>::type;
+    using pointer = typename std::add_pointer<
+        typename std::add_const<T>::type>::type;
     using reference = typename std::add_lvalue_reference<
         typename std::add_const<T>::type>::type;
+    using iterator_category = std::forward_iterator_tag;
 
     const T &operator*() const { return *static_cast<const T *>(this->pos); }
     const_iterator &operator++() {
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -243,10 +243,18 @@ public:
         typename std::add_const<T>::type>::type;
     using iterator_category = std::forward_iterator_tag;
 
+    const_iterator(): pos(nullptr), stride(0) { }
+    const_iterator(const void *pos, size_t stride): pos(pos), stride(stride) { }
     const T &operator*() const { return *static_cast<const T *>(this->pos); }
+    const T *operator->() const { return static_cast<const T *>(this->pos); }
     const_iterator &operator++() {
       this->pos = static_cast<const uint8_t *>(this->pos) + this->stride;
       return *this;
+    }
+    const_iterator operator++(int) {
+      auto ret = const_iterator(this->pos, this->stride);
+      this->pos = static_cast<const uint8_t *>(this->pos) + this->stride;
+      return ret;
     }
     bool operator==(const const_iterator &other) const {
       return this->pos == other.pos;
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -243,8 +243,6 @@ public:
         typename std::add_const<T>::type>::type;
     using iterator_category = std::forward_iterator_tag;
 
-    const_iterator(): pos(nullptr), stride(0) { }
-    const_iterator(const void *pos, size_t stride): pos(pos), stride(stride) { }
     const T &operator*() const { return *static_cast<const T *>(this->pos); }
     const T *operator->() const { return static_cast<const T *>(this->pos); }
     const_iterator &operator++() {
@@ -252,7 +250,7 @@ public:
       return *this;
     }
     const_iterator operator++(int) {
-      auto ret = const_iterator(this->pos, this->stride);
+      auto ret = const_iterator(*this);
       this->pos = static_cast<const uint8_t *>(this->pos) + this->stride;
       return ret;
     }
--- a/tests/ffi/tests.cc
+++ b/tests/ffi/tests.cc
@@ -200,6 +200,10 @@ void c_take_rust_vec_shared(rust::Vec<Shared> v) {
 }
 
 void c_take_ref_rust_vec(const rust::Vec<uint8_t> &v) {
+  // Test vector copy which was not compiling under gcc 7
+  std::vector<uint8_t> cxx_v;
+  std::copy(v.begin(), v.end(), back_inserter(cxx_v));
+
   uint8_t sum = std::accumulate(v.begin(), v.end(), 0);
   if (sum == 200) {
     cxx_test_suite_set_correct();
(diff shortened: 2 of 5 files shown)


myronahn:
Important note from the reference page for `std::iterator_traits`:

If Iter does not have all five member types difference_type, value_type, pointer, reference, and iterator_category, then this template has no members by any of those names (std::iterator_traits is SFINAE-friendly)

include/cxx.h
```diff
@@ -235,9 +235,13 @@ class Vec final {
 
   class const_iterator {
   public:
+    using difference_type = ptrdiff_t;
     using value_type = typename std::add_const<T>::type;
+    using pointer = typename std::add_pointer<
+        typename std::add_const<T>::type>::type;
     using reference = typename std::add_lvalue_reference<
         typename std::add_const<T>::type>::type;
+    using iterator_category = std::forward_iterator_tag;
```
dtolnay, line 244: I think this isn't quite correct because there are requirements on forward iterator not satisfied by this code yet.

https://en.cppreference.com/w/cpp/named_req/ForwardIterator

```cpp
void c_take_rust_vec_shared(rust::Vec<Shared> v) {
  auto it = v.begin();
  (void)it++;
  (void)it->z;
```

```console
tests.cc: In function ‘void tests::c_take_rust_vec_shared(rust::cxxbridge02::Vec<tests::Shared>)’:
tests.cc:195:11: error: no ‘operator++(int)’ declared for postfix ‘++’ [-fpermissive]
  195 |   (void)it++;
      |         ~~^~
tests.cc:196:11: error: base operand of ‘->’ has non-pointer type ‘rust::cxxbridge02::Vec<tests::Shared>::const_iterator’
  196 |   (void)it->z;
      |           ^~
```

tests/ffi/tests.cc
```diff
@@ -206,6 +206,16 @@ void c_take_ref_rust_vec(const rust::Vec<uint8_t> &v) {
...
 }
 
+void c_take_ref_rust_vec_copy(const rust::Vec<uint8_t> &v) {
+  // Test vector copy which was not compiling under gcc 7
+  std::vector<uint8_t> cxx_v;
+  std::copy(v.begin(), v.end(), back_inserter(cxx_v));
+  uint8_t sum = std::accumulate(cxx_v.begin(), cxx_v.end(), 0);
+  if (sum == 200) {
+    cxx_test_suite_set_correct();
+  }
```
dtolnay, line 216: Let's add a call from tests/test.rs so this assertion is actually executed.

tests/ffi/tests.cc
```diff
@@ -206,6 +206,16 @@ void c_take_ref_rust_vec(const rust::Vec<uint8_t> &v) {
   }
 }
 
+void c_take_ref_rust_vec_copy(const rust::Vec<uint8_t> &v) {
+  // Test vector copy which was not compiling under gcc 7
```
dtolnay, line 210: Please omit this comment, or change the comment to explain that this exercises the requirements of InputIterator.

tests/ffi/tests.cc
myronahn: Resolved in ba22ba8

tests/ffi/tests.cc
myronahn: Put in a better comment ae95e07

include/cxx.h
myronahn: Implemented forward iterator requirements c0fb2b0

include/cxx.h
```diff
@@ -235,15 +235,27 @@ class Vec final {
...
 
+    const_iterator(): pos(nullptr), stride(0) { }
+    const_iterator(const void *pos, size_t stride): pos(pos), stride(stride) { }
```
dtolnay, line 247: I would prefer not to offer this constructor. Only `friend class Vec` should be creating non-default iterators.

tests/ffi/tests.cc
```diff
@@ -203,13 +203,37 @@ void c_take_rust_vec_shared(rust::Vec<Shared> v) {
   }
 }
 
+void c_take_rust_vec_shared_forward_iterator(rust::Vec<Shared> v) {
+  // Exercise requirements of ForwardIterator
+  // https://en.cppreference.com/w/cpp/named_req/ForwardIterator
+  uint32_t sum = 0;
+  for (auto it = v.begin(); it != v.end(); it++) {
```
dtolnay, line 210: The ForwardIterator behavior is already covered by the previous test in this file which desugars to:

```cpp
for (auto __begin = v.begin(), __end = v.end(); __begin != __end; ++__begin) {
  /* ... */
}
```

I don't think it's worthwhile to keep an additional test like this, but if you think this is something people would write, we should make it call v.end() in the initializer so that anyone copying the code will avoid an extra 3 foreign calls on every iteration.

include/cxx.h
```diff
@@ -235,15 +235,27 @@ class Vec final {
...
   public:
+    using difference_type = ptrdiff_t;
     using value_type = typename std::add_const<T>::type;
+    using pointer = typename std::add_pointer<
+        typename std::add_const<T>::type>::type;
     using reference = typename std::add_lvalue_reference<
         typename std::add_const<T>::type>::type;
+    using iterator_category = std::forward_iterator_tag;
 
+    const_iterator(): pos(nullptr), stride(0) { }
```
dtolnay, line 246: We don't need to initialize the members because the default constructed iterator isn't required to have any particular behavior.

After deleting the other constructor, this class gets an implicitly declared default constructor which is sufficient for the purpose of forward_iterator, so this can be removed.

include/cxx.h
myronahn: Makes sense.  It made more sense with some code I had before, but now it is competely unnecessary.  Removed in 72ceca3.

include/cxx.h
myronahn: Makes sense.  Removed in 72ceca3.

tests/ffi/tests.cc
myronahn: This tests two things on the `const_iterator` that the other test doesn't cover: the postfix ++ operator and the the arrow operator.  

The arrow operator might not be a requirement for the forward iterator.  You mentioned it in your example so I included it:

```c++
void c_take_rust_vec_shared(rust::Vec<Shared> v) {
  auto it = v.begin();
  (void)it++;
  (void)it->z;
```

I went ahead and made the test more efficient in 7366d7f.

PR #159, opened 2020-04-27
Allow wildcard argument names.

### LINKED ISSUE #154: Allow wildcard argument names in extern function signature
Currently we reject this for no particular reason. Seems like an oversight.

```rust
#[cxx::bridge]
mod ffi {
    extern "C" {
        fn repro(_: &str);
    }
}
```

```console
error[cxxbridge]: unsupported signature

    ┌── src/main.rs:4:18 ───
    │
  4 │         fn repro(_: &str);  // but `s: &str` works
    │                  ^^^^^^^ unsupported signature
```

jgalenson:
This adds support for wildcard variable names by internally giving them unique names.

Fixes #154.

This seems simpler than trying to ensure we catch all uses of an ident named "_" and naming them appropriately.

### DIFF
--- a/syntax/parse.rs
+++ b/syntax/parse.rs
@@ -6,9 +6,9 @@ use crate::syntax::{
 use quote::{format_ident, quote};
 use syn::punctuated::Punctuated;
 use syn::{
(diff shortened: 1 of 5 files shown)


dtolnay:
Published in 0.2.12.

PR #169, opened 2020-04-30
Add support for rust::Vec<bool>

### LINKED ISSUE #152: Support Vec<bool>
C++ std::vector\<bool\> is [a little weird](https://en.cppreference.com/w/cpp/container/vector_bool) but there should be no problem with supporting Rust Vec\<bool\> at least; it should be exactly like all the other Rust vectors of primitives.

dtolnay:
Closes #152.

### DIFF
--- a/src/cxx.cc
+++ b/src/cxx.cc
@@ -266,7 +266,7 @@ void cxxbridge03$unique_ptr$std$string$drop(
   }
 
 // Usize and isize are the same type as one of the below.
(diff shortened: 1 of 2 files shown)


radekvit:
As of 1.0.76, trying to return `Vec<bool>` from an `extern "C++"` block doesn't work anymore.

Were there changes that deliberately disabled this support, or is this a bug?

Issue #132, opened 2020-04-22
Support C-style enums


jgalenson:
It would be nice to support C-style enums both from Rust to C++ and C++ to Rust.

The former is straightforward, but for the latter we must handle C++ code that passes values outside the expected range (which is illegal in Rust).  We can do this by auto-generating a conversion function that checks whether the value from C++ is one of the expected values and aborting otherwise.

The #[non_exhaustive] attribute is a possibility instead of a conversion function, but I believe it would not help at the moment (as both the definition and uses are in the same crate).

dtolnay:
I am interested in supporting this better. Currently my work codebase is `static_cast`ing to primitives and back, which works but can get pretty heavy on syntax.

#\[non_exhaustive\] isn't a direct solution even across crates because it's still UB for the enum to have any value other than the defined ones at runtime. For example the size of `#[non_exhaustive] enum E { Only }` is 0 bytes because there is only one possible runtime value.

jgalenson:
In that case, should we go with a conversion function (at least for now)?

dtolnay:
I think the most appropriate correspondence would be like this:

```cpp
enum class SmallPrime: uint16_t {
  Two = 2,
  Three = 3,
  Five = 5,
};
```

```rust
#[derive(Copy, Clone, PartialEq, Eq)]
#[repr(transparent)]
pub struct SmallPrime(u16);

#[allow(non_upper_case_globals)]
impl SmallPrime {
    pub const Two: Self = SmallPrime(2);
    pub const Three: Self = SmallPrime(3);
    pub const Five: Self = SmallPrime(5);
}
```

This works exactly like a nonexhaustive enum but matches the foreign representation.

```rust
fn demo(p: SmallPrime) {
    if p == SmallPrime::Five {
        /* ... */
    }

    match p {
        SmallPrime::Two => {}
        SmallPrime::Three => {}
        SmallPrime::Five => {}
        _other => {}
    }
}
```

dtolnay:
Implemented in #170.

PR #170, opened 2020-04-30
Support C-style enums

jgalenson:
This adds support for passing C-style enums between Rust and C++.

We use the Rust representation for enums suggested by dtolnay in #132.
Note that as this does not use real enums, Rust code cannot treat them
as normal enums, e.g., by converting them to integers.  But common
uses such as pattern matching remain unchanged.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -3,7 +3,7 @@ use crate::gen::{include, Opt};
 use crate::syntax::atom::Atom::{self, *};
 use crate::syntax::namespace::Namespace;
 use crate::syntax::symbol::Symbol;
-use crate::syntax::{mangle, Api, ExternFn, ExternType, Signature, Struct, Type, Types, Var};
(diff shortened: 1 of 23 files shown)


jgalenson:
Note that I decided to duplicate the logic for computing discriminants in check.rs and expand and check.rs to keep parsing close to the original syntax.  But we could of course compute discriminants there to avoid the duplication.

And as an aside the logic for parsing a variant is slightly ugly; I don't know if we prefer more nested matches instead.

jgalenson:
It looks like the tests are failing because nightly rustc gives a different error message.  Is there a way to configure it to change that test based on the compiler, maybe by modifying .travis.yml, or should we just delete that one test for now?

dtolnay:
Because of this line, we only run ui tests on nightly, so the expected error message files should match what is emitted by current nightly. I don't think maintaining additional expected outputs for different compilers would be productive.

https://github.com/dtolnay/cxx/blob/d776519a75a5e26e3259f8008ea228f83193cdf8/tests/compiletest.rs#L1

PR #182, opened 2020-05-01
Expose the enum value to Rust as a field

### LINKED ISSUE #178: Expose enum value as integer
#170 produces "enums" with a private integer representation, and public associated constants corresponding to the variants.

```rust
#[derive(Copy, Clone, PartialEq, Eq)]
#[repr(transparent)]
pub enum SmallPrime(u32);

#[allow(non_upper_case_globals)]
impl SmallPrime {
    pub const Two: Self = SmallPrime(2);
    pub const Three: Self = SmallPrime(3);
    pub const Five: Self = SmallPrime(6);
}
```

This behaves like an ordinary Rust enum in all ways except that it enforces a wildcard match arm when matched (which is deliberate) and that it cannot be cast using `as` to a primitive (which is unfortunate).

We should expose some way to get the primitive value. Maybe a method, maybe just a straightforward public field like `n.repr`.

jgalenson:
Fixes #178.

### DIFF
--- a/macro/src/expand.rs
+++ b/macro/src/expand.rs
@@ -137,14 +137,16 @@ fn expand_enum(enm: &Enum) -> TokenStream {
                 .unwrap_or_else(|| prev_discriminant.map_or(0, |n| n + 1));
             *prev_discriminant = Some(discriminant);
             Some(quote! {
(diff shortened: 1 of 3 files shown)


jgalenson:
I'm just using git, so don't worry about me.  Thanks for asking!

PR #187, opened 2020-05-04
Support enums defined in C++ code.

### LINKED ISSUE #177: Extern C-like enums
The enum support from #170 doesn't apply to interop with existing enums for which C++ is the source of truth.

I wonder if we could handle this use case as follows:

```rust
mod ffi {
    enum SmallPrime {
        Two = 2,
        Three = 3,
        Five = 5,
    }

    extern "C" {
        type SmallPrime;
    }
}
```

where we ask that the enum variants be provided just like in #170, but if the same type name is also an extern "C" type then we don't emit a C++ enum class definition, but instead emit static assertions on the C++ side asserting that the cxx bridge discriminants are all the same as the existing C++ enum's values.

jgalenson:
This allows listing an enum in an extern "C" block as well as in the shared block.  In this case, we do not emit the C++ declaration of the enum but instead emit static assertions that it has the values that we expect.

Do you know a good way to test the assertions?  The ui tests don't seem to compile the generated C++ files, so I'm not sure how to do it.

I'm also considering refactoring the discriminant generation code somewhere now that it's mostly duplicated in three different places.

Closes #177.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -5,7 +5,7 @@ use crate::syntax::namespace::Namespace;
 use crate::syntax::symbol::Symbol;
 use crate::syntax::{mangle, Api, Enum, ExternFn, ExternType, Signature, Struct, Type, Types, Var};
 use proc_macro2::Ident;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
 pub(super) fn gen(
     namespace: &Namespace,
@@ -46,7 +46,6 @@ pub(super) fn gen(
         }
     }
 
-    let mut cxx_types = HashSet::new();
     let mut methods_for_type = HashMap::new();
     for api in apis {
         if let Api::RustFunction(efn) = api {
@@ -56,9 +55,6 @@ pub(super) fn gen(
                     .or_insert_with(Vec::new)
                     .push(efn);
             }
-        }
-        if let Api::CxxType(enm) = api {
-            cxx_types.insert(&enm.ident);
         }
     }
 
@@ -70,7 +66,7 @@ pub(super) fn gen(
             }
             Api::Enum(enm) => {
                 out.next_section();
-                if cxx_types.contains(&enm.ident) {
+                if types.cxx.contains(&enm.ident) {
                     check_enum(out, enm);
                 } else {
                     write_enum(out, enm);
@@ -382,31 +378,24 @@ fn write_enum(out: &mut OutFile, enm: &Enum) {
 }
 
 fn check_enum(out: &mut OutFile, enm: &Enum) {
-    let discriminants = enm
-        .variants
-        .iter()
-        .scan(None, |prev_discriminant, variant| {
-            let discriminant = variant
-                .discriminant
-                .unwrap_or_else(|| prev_discriminant.map_or(0, |n| n + 1));
-            *prev_discriminant = Some(discriminant);
-            Some(discriminant)
-        });
     writeln!(
         out,
         "static_assert(sizeof({}) == sizeof(uint32_t));",
         enm.ident
     );
-    enm.variants
-        .iter()
-        .zip(discriminants)
-        .for_each(|(variant, discriminant)| {
-            writeln!(
-                out,
-                "static_assert({}::{} == {});",
-                enm.ident, variant.ident, discriminant
-            );
-        });
+    let mut prev_discriminant = None;
+    for variant in &enm.variants {
+        let discriminant = variant
+            .discriminant
+            .unwrap_or_else(|| prev_discriminant.map_or(0, |n| n + 1));
+        writeln!(
+            out,
+            "static_assert(static_cast<uint32_t>({}::{}) == {},
+              \"disagrees with the value in #[cxx::bridge]\");",
+            enm.ident, variant.ident, discriminant,
+        );
+        prev_discriminant = Some(discriminant);
+    }
 }
 
 fn write_exception_glue(out: &mut OutFile, apis: &[Api]) {
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -5,7 +5,7 @@ use crate::syntax::namespace::Namespace;
 use crate::syntax::symbol::Symbol;
 use crate::syntax::{mangle, Api, Enum, ExternFn, ExternType, Signature, Struct, Type, Types, Var};
 use proc_macro2::Ident;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 pub(super) fn gen(
     namespace: &Namespace,
@@ -46,6 +46,7 @@ pub(super) fn gen(
         }
     }
 
+    let mut cxx_types = HashSet::new();
     let mut methods_for_type = HashMap::new();
     for api in apis {
         if let Api::RustFunction(efn) = api {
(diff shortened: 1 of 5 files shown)


syntax/types.rs
```diff
@@ -69,7 +69,9 @@ impl<'a> Types<'a> {
                 }
                 Api::CxxType(ety) => {
                     let ident = &ety.ident;
-                    if !type_names.insert(ident) {
+                    // We allow declaring the same type as a shared enum and as a Cxxtype, as this
+                    // means not to emit the C++ enum definition.
+                    if !type_names.insert(ident) && !enums.contains_key(ident) {
```
dtolnay, line 74: I'd like for everything in Rust to stay order independent as much as possible, so please also deal with the Api::Enum being written after the Api::CxxType.

```rust
mod ffi {
    extern "C" {
        type E;
    }

    enum E {...}
}
```

It's fine to refactor how this duplication detection works if needed.

gen/src/write.rs
```diff
@@ -56,6 +57,9 @@ pub(super) fn gen(
                     .push(efn);
             }
         }
+        if let Api::CxxType(enm) = api {
```
dtolnay, line 60: `enm` should always refer to a crate::syntax::Enum. We should use `ety` for crate::syntax::ExternType.

gen/src/write.rs
```diff
@@ -46,6 +46,7 @@ pub(super) fn gen(
         }
     }
 
+    let mut cxx_types = HashSet::new();
```
dtolnay, line 49: The set of extern C++ type names is already available in `types.cxx`. Does that existing set work for what we need here?

macro/src/expand.rs
```diff
@@ -39,12 +40,22 @@ fn expand(namespace: &Namespace, ffi: ItemMod, apis: &[Api], types: &Types) -> T
         }
     }
 
+    let mut enums = HashSet::new();
+    for api in apis {
+        if let Api::Enum(enm) = api {
+            enums.insert(&enm.ident);
+        }
+    }
```
dtolnay, line 48: This set is already available if you use `types.enums.contains_key(...)`.

gen/src/write.rs
```diff
@@ -373,6 +381,34 @@ fn write_enum(out: &mut OutFile, enm: &Enum) {
...
+        "static_assert(sizeof({}) == sizeof(uint32_t));",
+        enm.ident
+    );
+    enm.variants
+        .iter()
+        .zip(discriminants)
+        .for_each(|(variant, discriminant)| {
+            writeln!(
+                out,
+                "static_assert({}::{} == {});",
```
dtolnay, line 406: Let's add a message argument here to support pre-C++17.
https://en.cppreference.com/w/cpp/language/static_assert
Something like `"disagrees with the value in #[cxx::bridge]"`.

dtolnay: Separately, it looks like this expands to an expression like `E::A == 0`? That didn't work on my machine.

```console
$ cat main.cc 
#include <cstdint>

enum class E : uint32_t {
  A = 0,
};

static_assert(E::A == 0);
```

```console
$ clang++ main.cc -std=c++17
main.cc:7:20: error: invalid operands to binary expression ('E' and 'int')
static_assert(E::A == 0);
              ~~~~ ^  ~
1 error generated.
```

```console
$ g++ main.cc -std=c++17
main.cc:7:20: error: no match for ‘operator==’ (operand types are ‘E’ and ‘int’)
    7 | static_assert(E::A == 0);
      |               ~~~~ ^~ ~
      |                  |    |
      |                  E    int
```

gen/src/write.rs
```diff
@@ -373,6 +381,34 @@ fn write_enum(out: &mut OutFile, enm: &Enum) {
...
+        });
+    writeln!(
+        out,
+        "static_assert(sizeof({}) == sizeof(uint32_t));",
+        enm.ident
+    );
+    enm.variants
+        .iter()
+        .zip(discriminants)
+        .for_each(|(variant, discriminant)| {
```
dtolnay, line 403: I feel like this went a bit overboard with scan/zip/for_each usage. I think this logic would be better suited to a loop.

```rust
let mut prev_discriminant = None;
for variant in &enm.variants {
    let discriminant = variant
        .discriminant
        .unwrap_or_else(|| prev_discriminant.map_or(0, |n| n + 1));
    writeln!(
        out,
        "static_assert({}::{} == {});",
        enm.ident, variant.ident, discriminant,
    );
    prev_discriminant = Some(discriminant);
}
```

Regarding your note in the PR description:

> I'm also considering refactoring the discriminant generation code somewhere now that it's mostly duplicated in three different places.

I agree. I think it would be reasonable to store a precomputed discriminant directly in crate::syntax::Variant (i.e. instead of an Option). Then:

```rust
for variant in &enm.variants {
    writeln!(
        out,
        "static_assert({}::{} == {});",
        enm.ident, variant.ident, variant.discriminant,
    );
}
```

syntax/types.rs
```diff
@@ -69,7 +69,9 @@ impl<'a> Types<'a> {
                 }
                 Api::CxxType(ety) => {
                     let ident = &ety.ident;
-                    if !type_names.insert(ident) {
+                    // We allow declaring the same type as a shared enum and as a Cxxtype, as this
+                    // means not to emit the C++ enum definition.
```
dtolnay, line 73: Please add a use of this to tests/ffi/lib.rs + tests/ffi/tests.h.

syntax/types.rs
jgalenson: Thanks for catching this!

gen/src/write.rs
jgalenson: Ah, I didn't see those.  Thanks!

gen/src/write.rs
jgalenson: Thanks for the C++17 tip.  I didn't know that.

About the miscompile, in my test code I was just doing "enum A" and not "enum class A", in which case it compiled.  But I added static_casts so it works both ways.  Thanks!

gen/src/write.rs
jgalenson: The reason I didn't compute the discriminant during parsing in my original patch was to keep it as close to the source as possible.  If you want to change that, we can.

I actually already had a patch refactoring this into a helper method, but it didn't remove the zips so you probably wouldn't like it. :)

dtolnay:
> The reason I didn't compute the discriminant during parsing in my original patch was to keep it as close to the source as possible. If you want to change that, we can.

I think the goal of the representation under crate::syntax::Api would be to be useful to the later compiler stages (check.rs, expand.rs + write.rs), not necessarily to be close to the source. That means normalizing all insignificant differences. For example:

- Whether the person typed `fn f(&self)` or `fn f(self: &Ident)` doesn't matter outside the parser, so those come out of the parser looking exactly the same.

- Similarly, whether they typed `extern "C" { fn f(); fn g(); }` or `extern "C" { fn f(); } extern "C" { fn g(); }` isn't significant, so those come out of the parser the same.

- I would want the same for `enum E { A, B }` vs `enum E { A = 0, B }` vs `enum E { A, B = 1 }` vs `enum E { A = 0, B = 1 }`. These are only different in ways that only matter to the parser.

jgalenson:
That makes sense.  I'll follow up with that tomorrow.

PR #191, opened 2020-05-08
bump cc to 1.0.52

philipcraig:


philipcraig:
> Is there a reason you need this? `cc = "1.0.49"` allows building with >=1.0.49 <2.0.0 so 1.0.52 should already work. Why do we need to drop support for the versions in between?

happy to drop it

PR #195, opened 2020-05-10
use make_unique in cpp example

philipcraig:
As it's more vernacular due to exception safety (no naked news)

### DIFF
--- a/demo-cxx/demo.cc
+++ b/demo-cxx/demo.cc
@@ -10,7 +10,7 @@ ThingC::ThingC(std::string appname) : appname(std::move(appname)) {}
 ThingC::~ThingC() { std::cout << "done with ThingC" << std::endl; }
 
 std::unique_ptr<ThingC> make_demo(rust::Str appname) {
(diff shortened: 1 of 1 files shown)


philipcraig:
Looks like this is failing as it needs C++14 or higher.

Seeing as I'm working on adding support for `std::optional`, which will require C++17 or higher, @dtolnay what restrictions or not would you like on C++ standard versions supported and used?

philipcraig:
> The demo code can use any standard, that doesn't matter.
> 
> The main library should be opt-in for particular standards (and other interop like absl) as discussed in #80.

Got it. I can see how to opt-in feature detection in the C++ code, like it has been done at https://github.com/dtolnay/cxx/blob/b01e5ac7a56eda0c91e68d4353e835b4cd45ba3c/include/cxx.h#L17-L36

However, I'm not sure how to get the knowledge that the user has deliberately opted-in to a C++17 feature (by e.g. using a `std::optional`) into the `build.rs` files to set the standard flag correctly for the C++ compilers.

Any thoughts on how this can be done?

PR #196, opened 2020-05-11
Choose enum discriminant type based on repr attributes, suffixes, values

### LINKED ISSUE #175: Represent enums using the smallest sufficient integer type
The C-style enum support added in #170 unconditionally uses `u32` / `uint32_t` as the representation. However, this means a negative discriminant like `-1` or a large discriminant like `0xfffffffff` won't compile. Additionally, an explicitly typed discriminant literal like `0u8` just ignores the type.

It would be nice to infer the intended discriminant type from the values. For example:

- If the discriminants are `0`, `1`, `2` -> `u8`
- If they are `-1`, `0`, `1` -> `i8`
- If they are `0u16`, `1`, `2` -> `u16`

dtolnay:
Fixes #175.

In order, the choice of repr is based on:

- **#\[repr(...)\] attribute**

    ```rust
    #[repr(u16)]
    enum E {  //:u16
        A,
        B,
    }
    ```

- otherwise, **integer suffix**

    ```rust
    enum E {  //:isize
        A = 0isize,
        B,
    }
    ```

- otherwise, **range of discriminant values**

    The first of `u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`, `i64` that can fit all of the values.

    ```rust
    enum E {  //:i16
        A = -1,
        B = 2020,
    }
    ```

FYI @jgalenson

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -365,7 +365,9 @@ fn write_enum(out: &mut OutFile, enm: &Enum) {
     for line in enm.doc.to_string().lines() {
         writeln!(out, "//{}", line);
     }
-    writeln!(out, "enum class {} : uint32_t {{", enm.ident);
+    write!(out, "enum class {} : ", enm.ident);
+    write_atom(out, enm.repr);
+    writeln!(out, " {{");
     for variant in &enm.variants {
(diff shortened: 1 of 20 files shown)


jgalenson:
Thanks!

PR #202, opened 2020-05-12
Add cargo cfgs for opt in to newer standards

dtolnay:
As described in https://github.com/dtolnay/cxx/pull/80#issuecomment-605794135. Currently unused, but will be needed when integrating std::optional or std::string_view support.

```toml
# Cargo.toml

[dependencies]
cxx = { version = "0.3", features = ["c++17"] }
```


philipcraig:
> FYI @philipcraig

awesome, thanks. Will need enhacement for MSVC when `std::optional` lands

PR #205, opened 2020-05-12
specify c++ standard flag for msvc

philipcraig:


philipcraig:
I can see that it's the four builds on nightly that are failing

PR #210, opened 2020-05-22
Fix issue with indirect return for C++ member function w/no args

myronahn:
A FFI C++ member function (i.e. function w/receiver) with no args and an indirect return (i.e. one that returns a `Result<T>`) was generating a function signature with a missing comma between the self arg and the indirect return parameter.

The `get_fail()` test member function will trigger this error.

The one-liner fix is in `gen/src/write.rs`


dtolnay:
Thanks, published in 0.3.4.

PR #214, opened 2020-05-31
Write generated bridge files to OUT_DIR as-is

### LINKED ISSUE #213: Cannot `#include gen.rs.h` reliably when building with cargo due to dynamic out location
Thanks for cxx!

It seems that the location of the generated bridge files varies based on the method with which the crate is added as dependency. E.g. in the crate [keytar](https://github.com/stoically/keytar-rs/tree/a0d5ec84940a2355a4a820ad94faa4934370b65b) I've used [`#include "src/lib.rs.h"`](https://github.com/stoically/keytar-rs/blob/a0d5ec84940a2355a4a820ad94faa4934370b65b/src/lib.cc#L2), which works fine while working directly in the crate, since the `rs.h` file ends up in `TARGET_DIR/cxxbridge/src/lib.rs.h`. When adding the crate as `dependency` (ends up in `TARGET_DIR/cxxbridge/.cargo/..`) or including it via `path` (ends up in nested paths depending on the local directory name), the out location changes in unpredictable ways, which makes it hard (impossible?) to reliably `#include` it.

My naive approach to fix this would be to always write the given include path as-is to the `OUT_DIR` and make it an requirement to include that (or automatically include it as part of `cxx_build::bridge`).

Maybe I'm missing something obvious and it is possible to reliably define `#include`s for cargo, in which case I'd appreciate pointers. Thanks!

stoically:
Possible way to fix #213. Currently breaks things, if this is something that might make sense I'm happy to look into making it work.

This requires consumers to add

```rust
let out_dir = env::var("OUT_DIR").unwrap();
// ...
cxx.include(out_dir);
```

to their build script.

### DIFF
--- a/gen/build/src/paths.rs
+++ b/gen/build/src/paths.rs
@@ -58,12 +58,11 @@ fn relative_to_parent_of_target_dir(original: &Path) -> Result<PathBuf> {
 }
 
 pub(crate) fn out_with_extension(path: &Path, ext: &str) -> Result<PathBuf> {
-    let mut file_name = path.file_name().unwrap().to_owned();
-    file_name.push(ext);
-
     let out_dir = out_dir()?;
-    let rel = relative_to_parent_of_target_dir(path)?;
-    Ok(out_dir.join(rel).with_file_name(file_name))
+    let mut out = out_dir.join(path).as_os_str().to_owned();
+    out.push(ext);
+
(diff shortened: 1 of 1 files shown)


stoically:
Works like a charm now. Thanks a lot for working on this! (And sorry that I didn't pick up my slack here after your fast response)

Issue #161, opened 2020-04-28
Ruleset for types


myronahn:
This is a half-baked idea to get some feedback as to whether you think it might be a good direction for the project.  It's a pretty big change so maybe it is best to discuss at a high level.

When I was implementing `CxxVector<T>` and `rust::Vec<T>`, one slightly tedious aspect was tracking down all the places where special handling was done for the types, e.g. searching for a `match` statement on the type and figuring out what the `match` statement was trying to achieve and then inserting a new match arm.  I think I ended up just searching the code for `RustBox` or `UniquePtr` and seeing if I needed to modify anything.  I was never 100% sure whether I actually handled all the special cases that I needed to. 

Perhaps we can define a trait (or an impl) with well-named functions that handle each special case.  For example, there might be a function to determine whether the type can be returned by value, whether it needs an indirect ABI, what is legal for its generic parameter `<T>`, and/or how to handle the type in the C++ and Rust function shim.  This would make it much easier to add new types.

myronahn:
Perhaps only worth doing if we feel that there are plenty more complex types that need to be passed across the FFI boundary.  Making this a turn-key operation would certainly encourage more types to be implemented for FFI.

myronahn:
This would give us a type-rule matrix.

When adding a new type, implement all the methods in the trait.

When adding a feature that requires a new rule, add that rule as a trait method and implement for all types (with sensible defaults).

jgalenson:
For what it's worth, I've played a bit with adding some new types and had both the same problem and same potential solution.  I'm not sure exactly how many types we think will be added, but something like this could make adding them a lot easier.

dtolnay:
I think this is worth trying. I put this off until we had a diverse set of supported types to generalize from, but we're just now at a point where I think future types will mostly be a combination of things we've done for the present ones, so this would be a good time to pursue it.

PR #216, opened 2020-06-09
WIP: Refactor C++ gen to centralize ruleset for types

myronahn:
This is a small beginning to #161 where the C++ generator is refactored such that there are very few match statements on types outside the traits `WriteType` and `WriteOptionType`.

Wanted to run it by everyone to see if this is a good direction in which to take the code.

The goal would be to have detailed comments on every trait function describing exactly where it is used.  The trait function names should also be descriptive (still working on that).

A similar trait would be defined for the Rust code generator macros.

### DIFF
--- a/gen/src/mod.rs
+++ b/gen/src/mod.rs
@@ -6,6 +6,7 @@ mod find;
 pub(super) mod include;
 pub(super) mod out;
 mod write;
+mod write_type;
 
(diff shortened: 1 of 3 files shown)


myronahn:
This is rapidly falling behind the current codebase.  I'll table it if there's no interest.

PR #220, opened 2020-06-13
cxx_name attribute - function aliases, import/export static methods, …

VKachemtsev:
I found a better solution for #218: cxx_side attribute
Features:
- Renaming functions/methods in C++ and Rust
- Import overloaded functions/methods
- Export/import static methods from both sides

In the future we can add new features for this attribute: replace return type to UniquePtr/Box, implement constructors, etc.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -575,8 +575,7 @@ fn write_function_pointer_trampoline(
 
     out.next_section();
     let c_trampoline = mangle::c_trampoline(&out.namespace, efn, var).to_string();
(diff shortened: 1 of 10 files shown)


dtolnay (commented):
Closing in favor of #218.

Thanks anyway!

jon-chuang:
Hi, currently it is impossible to do `#[cxx_name = "C::method"]` for static member function `method` of `C`. What is the workaround?

jon-chuang:
I guess the workaround is to wrap in non-static function. Thanks!

Issue #226, opened 2020-06-26
Warning on function pointers beginning nightly-2020-06-26


dtolnay:
A new warning emitted during `cargo test`:

```console
warning: `extern` fn uses type `fn(std::string::String) -> usize`, which is not FFI-safe
  --> tests/ffi/lib.rs:67:38
   |
67 |         fn c_take_callback(callback: fn(String) -> usize);
   |                                      ^^^^^^^^^^^^^^^^^^^ not FFI-safe
   |
   = note: `#[warn(improper_ctypes_definitions)]` on by default
   = help: consider using an `extern fn(...) -> ...` function pointer instead
   = note: this function pointer has Rust-specific calling convention
```

dtolnay:
Duplicate of #232.

PR #227, opened 2020-07-02
Delete symlinks in favor of relative paths

Bromeon:
Hello! When compiling cxx on Windows, I get several errors regarding unresolved modules:
```
error[E0583]: file not found for module `syntax`
  --> macro\src\lib.rs:13:1
   |
13 | mod syntax;
   | ^^^^^^^^^^^
   |
   = help: to create the module `syntax`, create file "macro\src\syntax.rs"

error[E0433]: failed to resolve: could not find `atom` in `syntax`
 --> macro\src\expand.rs:1:20
  |
1 | use crate::syntax::atom::Atom::{self, *};
  |                    ^^^^ could not find `atom` in `syntax`

(rest omitted)
```

When digging into it, it looks like you're using symlink files with relative paths, to refer to modules higher up.
Unfortunately, my Git for Windows client replaces symlinks with a textfile containing the relative path.

This PR replaces those with the filesystem-agnostic mechanism `#[path = ...] mod xy;`. The downside is that it needs to be done at each declaration site, but currently there are not many occurrences.

Let me know if there's a better way or if I missed some modules. Tests pass on latest Rust stable (not nightly due to #226).


Bromeon:
Seems to break the paths in Buck/Bazel builds, so might require more changes than I hoped for 😕

Bromeon:
Thanks for the feedback. I researched a bit, and the whole situation with symlinks on Windows [is unfortunately not easy](https://stackoverflow.com/q/5917249). There are various workarounds -- apparently for newer Git clients and explicit configuration, or with WSL + Regedit changes, it may work.

But exactly because this is so unreliable, I'm looking for an alternative that works without symlinks, ideally within Rust.

Some possible ways I was thinking of, also inspired by [this SO thread](https://stackoverflow.com/q/38417244):

* In the package step, copy the parent directories to the right place (build.rs?)
* Have different relative paths for the `package` step
* Restructure modules so this problem doesn't appear

But I also don't want to cause any bigger refactoring for this. Since I'm not very experienced with packaging, is there any other way you can think of? I assume similar problems arise every now and then when packaging more complex crates, are there any established best practices?

Issue #149, opened 2020-04-26
Fill out C++ API of rust::Vec<T>


dtolnay:
Our rust::Vec binding in C++ currently expose an extremely minimal API, basically just these:

```cpp
size_t size() const noexcept;
const T *data() const noexcept;
const_iterator begin() const noexcept;
const_iterator end() const noexcept;
```

It would be good to flesh this out some more, drawing inspiration from https://en.cppreference.com/w/cpp/container/vector. For example I would consider an `operator[]` pretty essential for a vector type.

dtolnay:
#261 expressed a need for push_back. I posted a workaround there but we really should have push_back built in.

sbrocket:
+1 push_back would be useful. A C++ to Rust move constructor would also be useful, i.e. `rust::Vec(std::vector&& other)`

dtolnay:
Update: cxx 0.5.9 added support for push_back and emplace_back.

https://github.com/dtolnay/cxx/blob/4ab117115718908768df8366ac130bbbb3478331/include/cxx.h#L180-L184

dtolnay:
https://cxx.rs/binding/vec.html

This is in fairly good shape now, so I'll go ahead and close this issue. We can file more specific follow-ups separately as people need particular APIs.

PR #257, opened 2020-08-21
Add rust::Vec accessors

rinon:
Adds operator[], at(), front(), and back() to rust::Vec.

This is the read-only accessor side of #149, not sure if we want mutable accessors.


dtolnay:
> not sure if we want mutable accessors.

Yes, I think we do want these!

PR #262, opened 2020-08-27
Unify the way that String construction and Vec indexing throw

dtolnay:
Following up on https://github.com/dtolnay/cxx/pull/257#pullrequestreview-476382515.

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -14,26 +14,6 @@
 #include <BaseTsd.h>
 #endif
 
-#ifndef __has_feature
-#define __has_feature(__x) 0
-#endif
-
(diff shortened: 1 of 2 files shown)


dtolnay:
MSVC doesn't like something. :(

```console
src/cxx.cc(295): error C2893: Failed to specialize function template 'void rust::cxxbridge03::panic(const char *)'
D:\a\cxx\cxx\src\../include/cxx.h(287): note: see declaration of 'rust::cxxbridge03::panic'
src/cxx.cc(295): note: With the following template arguments:
src/cxx.cc(295): note: 'Exception=std::out_of_range'
exit code: 2
```

Issue #297, opened 2020-09-11
Support for using shared types from one bridge in another bridge module


sbrocket:
I've been working on introducing cxx to Fuchsia as a safer, better way for us to write Rust/C++ FFIs. As part of this, I'd like to provide a library or two with some common types for others to reuse in their own FFIs.

For example, see the handle types from #274; these would be a bridge between our C++ and Rust handle wrappers. These are important to write once and reuse because there's a lot of ways to implement them in a way that would cause resource leaks. Another example is a reusable Result type; we don't use exceptions and panicking on every error result is undesirable.

To do this, I'd like to be able to reuse shared types defined in one bridge module in another bridge module without having to treat them as opaque Rust types in the using module (so that they can be passed by value without boxing). Today these seem to just be recognized as unsupported types. For example:

```
// handle.rs
#[cxx::bridge(namespace = zx::ffi)]
mod ffi {
    struct Process {
        raw: u32,
    }

    struct Job {
        raw: u32,
    }

    // ... and so on
}

impl Drop for ffi::Process { ... }
impl Drop for ffi::Job { ... }
// ... and so on, From impls to convert, etc.

// usage.rs
#[cxx::bridge(namespace = my_usage::ffi)]
mod ffi {
    extern "Rust" {
        fn create_process(name: &CxxString, job: &handle::ffi::Job) -> handle::ffi::Process;
    }
}
```

dtolnay:
This is not implemented yet but I am very interested in supporting it. We do something pretty close for https://docs.rs/cxx/0.4.4/cxx/trait.ExternType.html#safely-unifying-occurrences-of-the-same-extern-type but that's for sharing extern C++ types.

sbrocket:
Ok, I'm taking a crack at implementing it based on your PR #190

sbrocket:
One thing I noticed that don't seem to be clearly documented: The current ExternType support requires that two bridges use the same C++ namespace.

Is that necessary? It seems like this could be made to work across namespaces. I don't think the alias would need to explicitly specify the C++ namespace either, it seems like the ExternType trait could provide it. As long as you have the handshake of "the aliased type implements ExternType so we know it's a C++ type from another cxx bridge (or someone's manual unsafe impl)", I think that's all that needs to be verified? In other words, we just have to check that it's the right "flavor" of type, and then the using bridge needs to know the correct namespace to use when generating code. The type_id could be changed to only include the type name without the namespace, and a function added to ExternType return the namespace to use.

[...]

PR #298, opened 2020-09-13
Support type aliases across bridges using different C++ namespaces

sbrocket:
Related to #297


macro/src/expand.rs
```diff
@@ -667,8 +667,12 @@ fn expand_type_alias(alias: &TypeAlias) -> TokenStream {
 }
 
 fn expand_type_alias_verify(namespace: &Namespace, alias: &TypeAlias) -> TokenStream {
+    let namespace = alias.namespace.as_ref().unwrap_or(namespace);
+    // Review TODO: Is this unwrap fine? i.e. is it ok to assume that, if the
+    // TypePath parsed, that it has at least one segment?
```
sbrocket, line 672: Review question for you dtolnay@ (and there's a similar unwrap() in write.rs above)

tests/ffi/alias.rs
```diff
@@ -1,13 +1,20 @@
...
 #[rustfmt::skip]
-#[cxx::bridge(namespace = tests)]
+#[cxx::bridge(namespace = alias_tests)]
 pub mod ffi {
     extern "C" {
         include!("cxx-test-suite/tests.h");
 
+        // Review TODO: Unquoted namespace here doesn't work, is that expected or a bug
+        // in my parsing?
+        #[namespace = "tests"]
```
sbrocket, line 11: I'm new to proc-macros and usage of syn so it's unclear to me whether an unquoted namespace here should be able to work, as it does above.

```
error: unexpected token: `tests`
  --> tests/ffi/alias.rs:10:23
   |
6  |     extern "C" {
   |                - while parsing this item list starting here
...
10 |         #[namespace = tests]
   |                       ^^^^^
...
18 |     }
   |     - the item list ends here
```

macro/src/expand.rs
```diff
@@ -667,8 +667,12 @@ fn expand_type_alias(alias: &TypeAlias) -> TokenStream {
...
+    let remote_type = &alias.ty.path.segments.last().unwrap().ident;
```
sbrocket, line 673: Is this unwrap fine or is there a better way to handle this? (There's a similar one above in write.rs)

gen/src/write.rs
```diff
@@ -936,7 +936,10 @@ fn write_type(out: &mut OutFile, ty: &Type) {
 
 fn write_alias(out: &mut OutFile, alias: &TypeAlias) {
     if let Some(namespace) = &alias.namespace {
-        let path = namespace.path_for_type(&alias.ident);
+        // Review TODO: Is this unwrap fine? i.e. is it ok to assume that, if
+        // the TypePath parsed, that it has at least one segment?
+        let remote_type = &alias.ty.path.segments.last().unwrap().ident;
```
sbrocket, line 941: Is this unwrap (and the one below) fine or is there a better way to handle this?

tests/ffi/alias.rs
```diff
@@ -11,7 +11,10 @@ pub mod ffi {
         #[namespace = "tests"]
         type C = crate::ffi::C;
 
+        #[namespace = "tests"]
```
sbrocket, line 14: Needing a namespace attribute on every alias is going to get annoying for bridges with lots of aliases, but this at least works. An improved alternative might be to provide a mapping of Rust module to C++ namespace, e.g. "crate::ffi = tests", through the cxx::bridge attribute, but that's a bigger change that I'm less confident implementing right now.

macro/src/lib.rs
```diff
@@ -41,7 +41,10 @@ use syn::parse_macro_input;
 pub fn bridge(args: TokenStream, input: TokenStream) -> TokenStream {
     let _ = syntax::error::ERRORS;
 
-    let namespace = parse_macro_input!(args as Namespace);
+    let namespace = match Namespace::parse_bridge_attr_namespace.parse(args) {
+        Ok(ns) => ns,
+        Err(err) => return err.to_compile_error().into(),
+    };
```
sbrocket, line 47: Wrote https://github.com/dtolnay/syn/pull/896 to add a case to parse_macro_input! for this usage

tests/ffi/alias.rs
sbrocket: Ok, it wasn't a bigger change than I was confident making. 👍  Improved version implemented.

gen/src/write.rs
sbrocket: Obsolete

tests/ffi/alias.rs
sbrocket: Obsolete

tests/ffi/alias.rs
```diff
@@ -2,16 +2,12 @@
 // https://github.com/rust-lang/rustfmt/issues/4159
 #[rustfmt::skip]
 #[cxx::bridge(namespace = alias_tests)]
+#[cxx::alias_namespace(crate::ffi = tests)]
```
sbrocket, line 5: Needs documentation but please let me know what you think first.

dtolnay (commented):
Thanks for the PR! I have not gotten a chance to read the implementation yet but I am skimming the tests (tests/ffi/alias\*.rs) -- my sense is that the usage is going to feel pretty clumsy as a partial solution for "how do we allow one bridge to involve items from more than one namespace", which is the actual thing to solve, I think. If a C++ function in one namespace has an argument type in a different namespace (like alias::ffi::c_take_unique_ptr in the test suite), requiring to express that as two separate cxx::bridge modules with an extern type alias from one to the other and an alias_namespace attribute seems too convoluted.

This isn't fully designed but I wonder if allowing namespace controlled per extern block (inheriting from the namespace of the cxx::bridge if not specified on the extern block) would solve the same thing better.

```rust
#[cxx::bridge]
mod ffi {
    #[namespace = types]
    extern "C" {
        type SameC;
    }
    #[namespace = alias::tests]
    extern "C" {
        fn take_unique_ptr(c: UniquePtr<SameC>);
    }
}
```

Then, for the case where you do actually want items to be split across different cxx::bridge modules for Rust code organization reasons, the above just composes with the existing support for sharing a single Rust type across all different mentions of the opaque C++ type, by inserting `= path::to::SameC`.

Any thoughts on this in comparison to the approach in the PR?

---

Regarding your earlier comment about "unexpected token" error on an unquoted namespace: that isn't a proc macro or syn thing, just a silly rustc parsing limitation. [See this playground for a repro without any macro involved.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7eeec648c3d420cdab534dc2ae8b4871) I opened https://github.com/rust-lang/rust/pull/76734 with a fix.

sbrocket:
> Thanks for the PR! I have not gotten a chance to read the implementation yet but I am skimming the tests (tests/ffi/alias*.rs) -- my sense is that the usage is going to feel pretty clumsy as a partial solution for "how do we allow one bridge to involve items from more than one namespace", which is the actual thing to solve, I think. If a C++ function in one namespace has an argument type in a different namespace (like alias::ffi::c_take_unique_ptr in the test suite), requiring to express that as two separate cxx::bridge modules with an extern type alias from one to the other and an alias_namespace attribute seems too convoluted.

Great point. There's a couple different use cases being solved for here. I've been approaching this mostly with my use case in mind - where I'm going to use this for CXX shared types (once I add aliasing support for that), not C++ opaque types - but the use case you bring up is also totally valid.

I think `cxx::alias_namespace` is fine for the case where you have a bridge module for the other namespace anyway, as is the case with the current tests and `alias::ffi::c_take_unique_ptr`. I definitely would not want to force users to create an extra bridge just to use this attribute, however.

> This isn't fully designed but I wonder if allowing namespace controlled per extern block (inheriting from the namespace of the cxx::bridge if not specified on the extern block) would solve the same thing better.
> 
> ```rust
> #[cxx::bridge]
> mod ffi {
>     #[namespace = types]
>     extern "C" {
>         type SameC;
>     }
>     #[namespace = alias::tests]
>     extern "C" {
>         fn take_unique_ptr(c: UniquePtr<SameC>);
>     }
> }
> ```
> 
> Then, for the case where you do actually want items to be split across different cxx::bridge modules for Rust code organization reasons, the above just composes with the existing support for sharing a single Rust type across all different mentions of the opaque C++ type, by inserting `= path::to::SameC`.
> 
> Any thoughts on this in comparison to the approach in the PR?

We need to keep the module-level attribute that I currently have to support shared type aliases (once I add those, right after or in parallel with this PR), but supporting an attribute at the extern block level as well makes a lot of sense to me. I had forgotten that you can have multiple extern blocks. What do you think of the example below?

```rust
#[cxx::bridge(namespace = alias::tests)]
#[cxx::alias_namespace(other::ffi = foobar)
mod ffi {
    type Shared = other::ffi::Shared;

    #[namespace = types]
    extern "C" {
        type SameC;
    }

    extern "C" {
        fn convert(c: UniquePtr<SameC>) -> Shared;
    }
}
```

Also, potential alternative names for "cxx::alias_namespace": "cxx::module_namespace", "cxx::remote_namespace", "cxx::bridge_namespace", or simply "cxx::namespace"

FYI: You may not have noticed if you didn't dig into the individual commits, but I had originally added exactly that `#[namespace = foo]` attribute but on individual alias items, before refactoring it out to the current `cxx::alias_namespace` module-level attribute. I can just reintroduce that and shift it to the extern block level.

> 
> Regarding your earlier comment about "unexpected token" error on an unquoted namespace: that isn't a proc macro or syn thing, just a silly rustc parsing limitation. [See this playground for a repro without any macro involved.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0c282021dc2f63895950f1cfd7fe9488) I opened [rust-lang/rust#76734](https://github.com/rust-lang/rust/pull/76734) with a fix.

Ah great, thanks!

sbrocket:
Ping, looking for a response to proposal above.

dtolnay:
Sure thing -- I apologize about the holdup. The timing was unfortunate here since I was on vacation from the day before this PR was submitted until Tuesday of this week, and now I am working through an enormous review backlog in both https://github.com/rust-lang/rust and my own projects in addition to work.

sbrocket:
No problem, just wanted to get it back on your radar and agree on a design so I don't go through multiple implementation iterations unnecessarily.

dtolnay (commented):
> We need to keep the module-level attribute that I currently have to support shared type aliases (once I add those, right after or in parallel with this PR)

I would still very much prefer no module-level alias_namespace attribute. I think the mental model for knowing when to reach for it is too complicated. Is there any other way to express what you need?

From the code snippet in https://github.com/dtolnay/cxx/pull/298#issuecomment-693606432 you could do:

```rust
#[cxx::bridge]
mod ffi {
    #[namespace = foobar]
    use other::ffi::Shared;

    #[namespace = types]
    extern "C" {
        type SameC;
    }

    #[namespace = alias::tests]
    extern "C" {
        fn convert(c: UniquePtr<SameC>) -> Shared;
    }
}
```

so basically:

```diff
- #[alias_namespace(other::ffi = foobar)]
...
-     type Shared = other::ffi::Shared;
+     #[namespace = foobar]
+     use other::ffi::Shared;
...
-     type Shared1 = other::ffi::Shared1;
-     type Shared2 = other::ffi::Shared2;
+     #[namespace = foobar]
+     use other::ffi::{Shared1, Shared2};
...
-     type MyShared = other::ffi::Shared;
+     #[namespace = foobar]
+     use other::ffi::Shared as MyShared;
```

sbrocket:
I’d say it depends on whether you’re still thinking of namespace as a module level setting, as it is today and what I had in mind initially.

I notice above you shifted the bridge’s `alias::tests` namespace out of the cxx::bridge attribute and down to the extern C block, even though it does nothing there. Are you thinking of removing it from that attribute? If namespace remains a module level setting, then alias_namespace makes sense to me as a way to specify “this other Rust bridge module uses namespace X” to match the namespace in cxx::bridge. It would feel weird if we kept the namespace argument in cxx::bridge and then didn’t have a similar module-level attribute when composing bridges. I was imagining the other namespace attribute would be rarely used honestly, but I may be underestimating how often folks will use opaque C++ types from multiple namespaces in a single bridge?

Putting the namespace attribute at only the individual alias and extern block level is certainly the most flexible option if you have to pick one, but conceptually something seems off to me about it that I can’t put my finger on. It doesn’t seem like how namespaces are naturally used in C++, if that makes sense, but I could be wrong. Maybe it’s because it feels odd to have to repeat the attribute if I have a module with both some shared types and an extern C block that should use the same namespace? It could also just feel weird because this doesn’t have a clear C++ analogue; normally you’d never need to specify the namespace at both the type declaration and everywhere it’s included.

I also see from your example above that you seem to be favoring use decls rather than type aliases for pulling in shared types from other modules. Are you thinking of switching the current type alias support to that? I guess we’d have to without the module level attribute, to avoid needing lots of namespace attributes to pull in multiple shared types.

Should I drop these PRs and just let you implement, or are these useful to you?

dtolnay:
> I notice above you shifted the bridge’s alias::tests namespace out of the cxx::bridge attribute and down to the extern C block, even though it does nothing there. Are you thinking of removing it from that attribute?

`#[namespace = ...]` should be a hierarchical thing. If an item (fn, struct, enum, type, use) directly has a `namespace` attribute then that's the namespace for that item. Otherwise if the surrounding `extern` block (if any) has a `namespace` attribute then that's the namespace. Otherwise if the surrounding module has a `namespace` then we fall back to that. Otherwise assume `::`.

<br>

> I may be underestimating how often folks will use opaque C++ types from multiple namespaces in a single bridge?

I think this is the case. And it's not just about opaque C++ types, but all FFI items. It is not unusual for a function's argument types or return type to be from a different namespace as the function. I indicated in https://github.com/dtolnay/cxx/pull/298#pullrequestreview-489364768 that I think the actual problem to solve is how to allow one bridge to involve items from more than one namespace, after which a resolution for type aliases across bridges would fall out naturally.

<br>

> It doesn’t seem like how namespaces are naturally used in C++, if that makes sense, but I could be wrong.

I would say I don't see this as a design goal. The way to view a cxx::bridge module is just as *describing* what exists in Rust and C++ that each side needs to know about the other. What matters to me is that the description is straightforward such that the individual pieces of functionality implemented by cxx::bridge are easily comprehensible in isolation and compose intuitively, not that it necessarily resembles semantics of C++ as closely as possible.

<br>

> I also see from your example above that you seem to be favoring `use` decls rather than type aliases for pulling in shared types from other modules. Are you thinking of switching the current type alias support to that?

Yes; the ExternType::Kind associate type/marker types from #325 make this possible because we now have a way for a downstream module to reflect on whether a particular symbol from outside the module is an opaque (extern) type or one which both Rust and C++ can hold by value.

<br>

> Should I drop these PRs and just let you implement, or are these useful to you?

The advantage you have is potentially more time able to dedicate to cxx, since I am also on the hook for rust-lang/rust and 100 other projects maintenance at the same time.

One thing that worked well in #325 was a substantial amount of back-and-forth in #313 to iterate and exactly pin down the design beforehand, such that the PR was implementing a design that we both already agreed was good.

For https://github.com/dtolnay/cxx/pull/298#pullrequestreview-501654503 there is a chain of reasonable implementation steps which I'd be working through at whatever rate I'm able, but if a PR exists, I would prioritize accepting that.

- Emit all mentions of a type in generated C++ code (in function arguments and so forth) as a fully qualified path `::name::space::of::Type`, even if it currently always happens to be within the same module.

- Store a distinct Namespace on each FFI item rather than having a single global namespace for the whole code generation.

- Implement `#[namespace = ...]` attribute to determine namespace of individual items.

- Implement hierarchical handling of namespace which prioritizes item-level attribute \> extern block \> module.

- Emit ExternType impls for shared structs.

- Safe ExternType derive for extern Rust items. `extern "Rust" { #[derive(ExternType)] type TheType; }`

- Permit `use` inside of cxx::bridge.

dtolnay (commented):
I've copied the implementation plan from the previous comment to #353, so I'll close the PR in favor of tracking it there.

Thanks anyway!

sbrocket:
I’ve been on vacation this week. Will read above in more detail and see if there’s ways I can contribute next week when I’m back.

PR #301, opened 2020-09-16
Make sure MaybeUninit is part of generated code for Rust functions returning Result

Nehliin:
The `indirect_return` functions checks if a function signature either throws or needs indirect ABI and decides if `::rust::MaybeUninit` should be added to the generated C++ code. However the declaration for the `::rust::MaybeUninit` was only generated if the function needed indirect ABI so I ended up with functions that returned a Result generated code that used `::rust::MaybeUninit` but didn't have the declaration. This PR should make sure that the `::rust::MaybeUninit` declaration is part of the generated code if a function returns a Result.


dtolnay:
Published in 0.4.5.

PR #310, opened 2020-09-21
Fix return Result<Box<T>> from Rust to C++

### LINKED ISSUE #309: Can't return Result<Box<T>>
I am trying to return a Result<Box< T >> and I am getting the below compile errors. 
>expected struct `std::boxed::Box`, found enum `std::result::Result`
>expected enum `std::result::Result`, found *-ptr

I have tried setting the return type to Result<T> and Box<T> and both compile fine.

Here is a simple proof. 

```rust
#[cxx::bridge()]
mod test {
    struct Test {
        size: usize,
    }

    extern "Rust" {
        fn test() -> Result<Box<Test>>;
    }   
}       
        
fn test() -> Result<Box<test::Test>> {
    Ok(Box::<test::Test>::new(test::Test{size: 0}))
}
```

The reason I want to return a Result<Box<T>> is because in my use case T is a rust opaque type, but I made T a shared type to check that both Result<T> and Box<T> compile fine where Result<Box<T>> doesn't.

dtolnay:
Fixes #309.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -775,9 +775,11 @@ fn write_rust_function_shim_impl(
         write!(out, "extern$");
     }
     write!(out, ")");
-    if let Some(ret) = &sig.ret {
-        if let Type::RustBox(_) | Type::UniquePtr(_) = ret {
-            write!(out, ")");
+    if !indirect_return {
+        if let Some(ret) = &sig.ret {
+            if let Type::RustBox(_) | Type::UniquePtr(_) = ret {
+                write!(out, ")");
+            }
         }
     }
     writeln!(out, ";");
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
Windows CI doesn't like the branch name. :rofl: 

```console
 Resolving deltas: 100% (8/8), done.
error: cannot lock ref 'refs/remotes/origin/Result<Box<T>>': Unable to create 'D:/a/cxx/cxx/.git/refs/remotes/origin/Result<Box<T>>.lock': Invalid argument
From https://github.com/dtolnay/cxx
 ! [new ref]         b86c71d4e666aecb6c22f60902f31ade3e6a689a -> origin/Result<Box<T>>  (unable to update local ref)
```

Issue #320, opened 2020-09-24
Retrieve slice from CxxVector


roligugus:
There is currently no way to get a slice out of a CxxVector. Let's add the a new get_slice() helper so that rust code can use a slice. 

Background, we're doing some C++ -> rust API in-house and there's been a wish for that functionality to have "nicer C++ APIs".

Yes, one can use a slice in the FFI API as a work-around. I.e. use rust::Slice<T> in C++.

Pull request incoming.

PR #321, opened 2020-09-24
Add API to get slice from CxxVector

roligugus:
Implementing #320 

Adds a get_slice() to CxxVector so that one can use slices in the rust code and std::vector on the C++ side.

Bear with me, new to rust.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -1272,6 +1272,13 @@ fn write_cxx_vector(out: &mut OutFile, vector_ty: &Type, element: &Ident, types:
     writeln!(out, "}}");
     writeln!(
         out,
+        "const {} *cxxbridge04$std$vector${}$data(const ::std::vector<{}> &s) noexcept {{",
+        inner, instance, inner,
+    );
+    writeln!(out, "  return s.data();");
+    writeln!(out, "}}");
+    writeln!(
+        out,
         "const {} *cxxbridge04$std$vector${}$get_unchecked(const ::std::vector<{}> &s, size_t pos) noexcept {{",
         inner, instance, inner,
     );
--- a/macro/src/expand.rs
+++ b/macro/src/expand.rs
@@ -867,8 +867,22 @@ fn expand_cxx_vector(namespace: &Namespace, elem: &Ident) -> TokenStream {
                     #[link_name = #link_data]
                     fn __vector_data(_: &::cxx::CxxVector<#elem>) -> *const #elem;
                 }
-                let ret = unsafe { std::slice::from_raw_parts(__vector_data(v), __vector_size(v)) };
-                ret
+                let data = unsafe { __vector_data(v) };
+                if (data.is_null())
+                {
+                    // Avoid undefined behaviour in slice::from_raw_parts() as
+                    // std::vector::data() returns nullptr if no capacity().
+                    // Alternative: Use from_raw_parts(NonNull::dangling(), 0)
+                    let ret = <&[Self]>::default(); // empty slice
+                    ret
+                }
+                else
+                {
+                    // std::vector with capacity() has valid data ptr, even if
+                    // size() == 0.
+                    let ret = unsafe { std::slice::from_raw_parts(data, __vector_size(v)) };
+                    ret
+                }
             }
             unsafe fn __get_unchecked(v: &::cxx::CxxVector<Self>, pos: usize) -> &Self {
                 extern "C" {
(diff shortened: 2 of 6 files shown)


gen/src/write.rs
```diff
@@ -1270,6 +1270,13 @@ fn write_cxx_vector(out: &mut OutFile, vector_ty: &Type, element: &Ident, types:
     );
     writeln!(out, "  return s.size();");
     writeln!(out, "}}");
+    writeln!(
```
roligugus, line 1273: Wondering whether I can or should generate src/cxx.h|cc from this. Need to look into that...

tests/ffi/tests.cc
```diff
@@ -401,6 +401,12 @@ extern "C" const char *cxx_run_test() noexcept {
       std::unique_ptr<std::string>(new std::string("2020")));
   r_take_enum(Enum::AVal);
 
+  // Caller owns data handed into rust APIs
```
roligugus, line 404: Test for that did not really fit into the "take" or "return" tests. 

I initially had a c_take...() test that would create a std::vector and then call into the r_access_...() test function, but I found that overly complex as the c_take...() test did not add anything. Hence, directly testing the API in r_access..() here.

macro/src/expand.rs
```diff
@@ -859,6 +860,16 @@ fn expand_cxx_vector(namespace: &Namespace, elem: &Ident) -> TokenStream {
...
                 unsafe { __vector_size(v) }
             }
+            fn __vector_slice(v: &::cxx::CxxVector<Self>) -> &[Self] {
+                extern "C" {
+                    #[link_name = #link_size]
+                    fn __vector_size(_: &::cxx::CxxVector<#elem>) -> usize;
+                    #[link_name = #link_data]
+                    fn __vector_data(_: &::cxx::CxxVector<#elem>) -> *const #elem;
+                }
+                let ret = unsafe { std::slice::from_raw_parts(__vector_data(v), __vector_size(v)) };
```
dtolnay, line 870: I believe this is undefined behavior in the case that size is 0. The data ptr of a &\[T\] has to be nonnull and properly aligned regardless of size, which C++ does not guarantee.

See https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety.

Somewhere there will need to be special handling of the empty vector case.

dtolnay (changes requested):
Thanks for the PR!

macro/src/expand.rs
roligugus: I was actually wondering about that. ;)
Let me add the special casing of nullptr/0 size

roligugus: Have not looked into how rust handles alignment. That could be a biggie if rust has different alignment reqs...

Having said that, we could enforce that in the __vector_data() implementation.

roligugus: Fixed handling of null data() ptr.

UB only happens if std::vector's capacity() is 0 if I got the from_raw_parts documentation right. There does not seem to be a need to specifically handle 0 size. Famous last words?

Ignore my going off  the tangent with non-alignment. I was thinking off T, whereas you were talking about the data ptr if null.

src/cxx_vector.rs
```diff
@@ -145,6 +157,34 @@ macro_rules! impl_vector_element {
...
+                    // std::vector::data() returns nullptr if no capacity().
+                    // Alternative: Use from_raw_parts(NonNull::dangling(), 0)
+                    let ret = <&[$ty]>::default(); // empty slice
+                    ret
+                }
+                else
+                {
+                    // std::vector with capacity() has valid data ptr, even if
+                    // size() == 0.
+                    let ret = unsafe { slice::from_raw_parts(data, __vector_size(v)) };
```
dtolnay, line 184: This is still undefined behavior, I think. As far as I can tell nothing in https://en.cppreference.com/w/cpp/container/vector/data would guarantee that `data` produces a sufficiently aligned pointer in the empty case. Can you find that in the standard, or implement a different way?

dtolnay (changes requested):

dtolnay (commented):
I ended up landing #322 with an implementation that I think is correct, and released in 0.4.7.

Thanks anyway!

src/cxx_vector.rs
roligugus: You are right. Went after the implementation behaviour that always has been instead of reading the standard.

Fix would be to check the size() == 0 instead the ptr, which I see you have done in your implementation. Thanks for fixing my code up. ;)

Issue #323, opened 2020-09-25
Demo: Use same C++ standard as cxxbridge


roligugus:
The demo code is currently set up to set the C++ standard by setting it in build.rs a la

```
    cxx_build::bridge("src/main.rs")
        .file("src/demo.cc")
        .flag_if_supported("-std=c++14")
        .compile("cxxbridge-demo");
```

However, cxx.h|cc are compiled with the CXX crate feature settings, i.e. cxxbridge_flags::STD. ffi tests already use the feature flag. 

We shall align the demo's build.rs to use cxxbridge_flags::STD so that we only have to declare the feature in demo's Cargo.toml.

While we're at it, it would be great to not having to set `flag_if_supported(cxxbridge_flags::STD)` when using cxx_build. cxx_build could directly set the feature flags. This makes the cxx_build usage less error-prone for cargo-based builds as it automatically ensures that custom FFI code and cxx.h|cc is compiled with the same features. Especially once we add C++17 features such as string_view support, or any other potential future feature such as absl, folly, or gsl support.

This is not an issue for bazel or buck builds.

Pull request incoming.

roligugus:
@dtolnay Just to double check. Was expecting to see this issue closed with "not at this time" due to the outcome of PR #324.

Oversight, or would you like a pull request that does the change in demo's build.rs to?
```
 cxx_build::bridge("src/main.rs")
        .file("src/demo.cc")
        .flag_if_supported(cxxbridge_flags::STD)
        .compile("cxxbridge-demo");
```

dtolnay:
I would prefer not to do this at this time, so I'll go ahead and close the issue. Thanks anyway!

PR #324, opened 2020-09-25
Automatically pick up cxx crate features in cxx_build

roligugus:
Fix and implementation for #323 

Let's make sure that the core lib (cxx.h|cc) and the demo code is compiled with the same features, i.e. C++ standard only at this time. These are the first two commits (code + documentation update):
- b780338 Set C++ std in Cargo.toml for demo
- 1fa8579 Align documentation to use cxxbridge_flags.STD with cxx_build

I also moved the setting of features to cxx_build to avoid any potential miscompilation between core (cxx.h|cc) and any cargo-based client compilation using cxx_build. cxx_build can already set the CXX crate's features for all client code. This enables clients to define the features required on the CXX crate in Cargo.toml and not having to think about adding the feature settings to their compilation in build.rs. Well, at least for cargo-based clients. Buck and bazel builds are different.

This should make adding new feature easier as well as one only needs to update cxx_build and then clients can enable the features in their Cargo.toml. This should help with C++17 (string_view) support as well as folly support that I will be looking into.

Moving the feature setting into cxx_build are commits 3&4:
- d328816 Set C++ std in cxx_build for simpler usage
- 4602329 Align documentation to no longer set cxxbridge_flags manually

Hope my explanation makes sense.

Anyhow, food for thought. Review best done by commit.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -51,6 +51,11 @@ rust_library(
 )
 
 rust_library(
+    name = "cxxbridge-flags",
+    srcs = glob(["flags/src/*.rs"]),
+)
+
+rust_library(
     name = "build",
     srcs = glob(["gen/build/src/**/*.rs"]),
     data = ["gen/build/src/gen/include/cxx.h"],
@@ -61,6 +66,7 @@ rust_library(
         "//third-party:proc-macro2",
         "//third-party:quote",
         "//third-party:syn",
+        "//:cxxbridge-flags",
     ],
 )
 
--- a/demo/Cargo.toml
+++ b/demo/Cargo.toml
@@ -10,5 +10,4 @@ cxx = { path = "..", features= ["c++14"] }
 
 [build-dependencies]
 cxx-build = { path = "../gen/build" }
-cxxbridge-flags = { path = "../flags" }
 
--- a/demo/Cargo.toml
+++ b/demo/Cargo.toml
@@ -6,7 +6,9 @@ edition = "2018"
 publish = false
 
 [dependencies]
-cxx = { path = ".." }
+cxx = { path = "..", features= ["c++14"] }
 
 [build-dependencies]
 cxx-build = { path = "../gen/build" }
+cxxbridge-flags = { path = "../flags" }
+
--- a/gen/build/src/lib.rs
+++ b/gen/build/src/lib.rs
@@ -4,6 +4,11 @@
 //! C++ code generator, set up any additional compiler flags depending on
 //! the use case, and make the C++ compiler invocation.
 //!
+//! Set CXX features such as support of a different C++ standard - default is
+//! C++11 - by setting the feature on the CXX crate. This ensures that the CXX
+//! bridge library code in cxx.h|cc gets compiled with the same feature flags
+//! as your code.
+//!
 //! <br>
 //!
 //! # Example
@@ -16,7 +21,7 @@
 //! fn main() {
 //!     cxx_build::bridge("src/main.rs")
 //!         .file("src/demo.cc")
-//!         .flag_if_supported("-std=c++11")
+//!         .flag_if_supported(cxxbridge_flags::STD)
 //!         .compile("cxxbridge-demo");
 //!
 //!     println!("cargo:rerun-if-changed=src/main.rs");
@@ -87,7 +92,7 @@ pub fn bridge(rust_source_file: impl AsRef<Path>) -> Build {
 /// let source_files = vec!["src/main.rs", "src/path/to/other.rs"];
 /// cxx_build::bridges(source_files)
 ///     .file("src/demo.cc")
-///     .flag_if_supported("-std=c++11")
+///     .flag_if_supported(cxxbridge_flags::STD)
 ///     .compile("cxxbridge-demo");
 /// ```
 #[must_use]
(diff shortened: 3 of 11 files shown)


BUILD
```diff
@@ -50,6 +50,11 @@ rust_library(
     ],
 )
 
+rust_library(
```
roligugus, line 53: cxx-build now depends on cxxbridge-flags and I had to add this to the bazel build.
Curiously enough, the buck build was ok?

Please note, I have no bazel experience, so this might be wrong/incomplete/...

demo/Cargo.toml
```diff
@@ -6,7 +6,8 @@ edition = "2018"
 publish = false
 
 [dependencies]
-cxx = { path = ".." }
+cxx = { path = "..", features= ["c++14"] }
```
roligugus, line 9: The C++ version for demo is now declared in Cargo.toml

gen/build/src/lib.rs
```diff
@@ -129,6 +132,8 @@ fn build(rust_source_files: &mut dyn Iterator<Item = impl AsRef<Path>>) -> Resul
     build.cpp(true);
     build.cpp_link_stdlib(None); // linked via link-cplusplus crate
     build.include(&include_dir);
+    // Set feature parameters, such as C++ std, for simpler usage
+    build.flag_if_supported(cxxbridge_flags::STD);
```
roligugus, line 136: In the future, we can add other feature, such as folly or absl support. cargo-build clients only will have to chose the feature in their Cargo.toml.

dtolnay (commented):
Is this necessary? As I understand it, it isn't a requirement that the standard be globally consistent across translation units. What potential miscompilation are you referring to?

roligugus:
> 
> 
> Is this necessary? As I understand it, it isn't a requirement that the standard be globally consistent across translation units. What potential miscompilation are you referring to?

E.g. only enabling c++17 in demo, but compiling cxx.h|cc with C++11 could result in the string_view API enabled while compiling the demo, but not when compiling and linking the core lib (cxx.h|cc).

dtolnay:
I wouldn't consider that a miscompilation. If the demo doesn't use string_view then there should be no problem. If the demo does use string_view then it needs to enable the corresponding feature on its dependency on cxx according to what standard types it needs.

roligugus:
Sorry was rushing this out before picking up kids at school yesterday...

> Is this necessary?

Not strictly. It's a usability/convenience functionality, so that I don't have to enable a feature in build.rs as well as on the cxx crate, but only in one place - the cxx crate. It theoretically also makes for less explaining/documentation as one only has to enable features on the cxx crate for cxx_build-based integration.

> As I understand it, it isn't a requirement that the standard be globally consistent across translation units.

That is correct. You'll be fine mixing it across TUs if compiled on the same machine with the same compiler. Same ABI for libstdc++/libc++. I might have some questions about using STL features only available in newer C++ releases when compiling with old standard as it pertains to platform behaviour, but the cxx usage should not run into these corners of STL usage. 

> What potential miscompilation are you referring to?

Sorry, I should have been more specific. Miscompilation to me is where you compile the core (cxx.h|cc) with different features/build settings than demo (that includes cxx.h). Problems happens mostly when you get into the game of enabling/disabling functionality in c++ libraries/headers with defines and where you end up having diverging APIs.

The result *might* be linking errors in some cases.

By setting compile options in demo's build.rs, demo will always compile fine. By setting features on the cxx crate, the core compilation will always be fine. However, where you can run into linking errors if you are using APIs in demo, that were not available in the core compilation. You might luck out on inlined APIs in cxx.h. Yes, there is no issue with using fewer features in demo than provided in the core if these APIs are additive as you mention.

Having said that, you *might* see a linking error and then you might react as I did when playing around with folly-based APIs enabled through defines: "Oh, I set that in build.rs. Ah, I also need to set it for cxx core. Would be nice to set it only one as I always will be compiling cxx core with the same features as my demo/real program.".

I personally find this use case to match features in cxx core and demo/real program more likely as upon building with cxx_build I will build the cxx core at the same time as demo/real program.

Is that a big deal? No. But I think it would be nicer to only having to set that in one place, i.e. the cxx crate as we can easily pick up the demo crate's cxx features in the cxx_build implementation.

dtolnay (commented):
I think I would prefer not to make this change.

Of the things you wrote, I definitely disagree with this part:

> It theoretically makes for less explaining/documentation as one only has to enable features on the cxx crate for cxx_build-based integration.

We generally expect people to include non-cxx-generated code in the same compiler invocation from build.rs, like the demo does with `src/demo.cc`, or could be a larger C++ library that they are binding against. If their non-cxx C++ code needs a recent standard, it's counterintuitive and requires *more* explanation that they can't just use the usual cc crate mechanisms for configuring the build, instead they need to go to Cargo.toml to set it as a cxx crate feature, despite not needing any of the support for "new" types (string_view etc) from cxx which that feature is ostensibly for.

For now I am open to waiting to see whether any specific build failures arise in practice, and evaluate how those are best addressed.

> However, where you can run into linking errors if you are using APIs in demo, that were not available in the core compilation. 

This is something we can catch at code generation time. If a cxx::bridge contains types that are not enabled on the cxx crate, we can immediately report a build error rather than continuing until link.

Issue #367, opened 2020-10-21
include! uses relative path, which breaks for dependencies


Bromeon:
Let's assume project `cxx-example` has a bridge module, which sets up the C++ bindings. I took it mostly from the `cxx-demo` crate inside the main repo.
```rust
#[cxx::bridge(namespace = org::example)]
mod ffi {
    extern "C" {
        include!("include/demo.h");

        ...
    }
}
```

The `include!` is translated to verbatim `#include "include/demo.h` in C++. Which file this refers to, depends on the include paths on the `cc::Build` configuration in `build.rs`.

If I make _another_ project `cxx-depend` which uses `cxx-example` as a dependency, then it fails with:

```
   Compiling cxx-example v0.1.0 (...\cxx-example)
error: failed to run custom build command for `cxx-example v0.1.0 (...\cxx-example)`

Caused by:
  process didn't exit successfully: `...\my_project\target\debug\build\cxx-example-6c41497ba3fecf2f\build-script-build` (exit code: 1)
  --- stderr

  CXX include path:
    ...\my_project\target\debug\build\cxx-example-2606dc5ab30a72a7\out\cxxbridge\include
    ...\my_project\target\debug\build\cxx-example-2606dc5ab30a72a7\out\cxxbridge\crate

  error occurred: Command 
"C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Community\\VC\\Tools\\MSVC\\14.27.29110\\bin\\HostX64\\x64\\cl.exe" 
  "-nologo" "-MD" "-Z7" "-Brepro" 
  "-I" "...\\my_project\\target\\debug\\build\\cxx-example-2606dc5ab30a72a7\\out\\cxxbridge\\include" 
  "-I" "...\\my_project\\target\\debug\\build\\cxx-example-2606dc5ab30a72a7\\out\\cxxbridge\\crate" 
  "-I" "...\\cxx-example" 
  "-W4"
  "-FoG:\\Rust\\Experiments\\my_project\\target\\debug\\build\\cxx-example-2606dc5ab30a72a7\\out\\src/demo.o" 
  "-c" "src/demo.cc"
```

Note that it fails at `cxx-example` (not the current crate `cxx-depend`), which I just successfully compiled.

In the error output, you see that the include paths refer to the current crate, although some included files (e.g. `cxx-example/include/demo.h`) are in the source tree of `cxx-example`. It seems like `include!(...)` does not take that into account.

[...]

PR #368, opened 2020-10-23
Add error message for invalid include! paths

Bromeon:
As promised in [#367](https://github.com/dtolnay/cxx/issues/367#issuecomment-714738184), here's the pull request for an error message when `include!` receives a path that it cannot handle. This is to avoid that such errors propagate further down and only fail at the C++ compile step (and was a reason I spend a lot of time debugging my cxx examples).

**The logic is:** path must begin with the configured include prefix (crate name by default) or must be an absolute path.

Remarks:
1. There were multiple places where I could have added the verification, but I thought the parser is not a good idea, because what `include!` means depends on the user of the parser (can be an external crate such as autocxx). So I implemented it close to the bridge generation.
1. In the parser, I now no longer skip `include!` statements, but integrate them with the `Types` struct and carry that information upward several levels.
1. There are unit tests for the `verify_include_dir()` function, but I'm not sure what's the best location for them. I cannot move them to another crate without making the function under test public.
1. For a similar reason I couldn't write any integration tests (testing the whole chain like in `tests/cxx_gen.rs`). I did some manual tests however. Probably you know much better how and where to add extra tests, if it's even necessary (this PR is primarily a quality-of-life/diagnostic improvement)

### DIFF
--- a/gen/build/src/lib.rs
+++ b/gen/build/src/lib.rs
@@ -311,8 +311,9 @@ fn verify_include_dir(include_dir: &PathBuf, include_prefix: &Path) {
         return;
     }
 
+    // Include prefix is stored with backslashes under Windows -> very basic slash conversion (use path-slash for more sophisticated one)
     panic!("include!(\"{}\") is invalid:\n   path must either start with include prefix '{}' or be an absolute path.\n\n",
-           include_dir.display(), include_prefix.display());
+           include_dir.display(), include_prefix.to_str().unwrap().to_string().replace("\\", "/"));
 }
 
 fn env_os(key: impl AsRef<OsStr>) -> Result<OsString> {
--- a/gen/build/src/lib.rs
+++ b/gen/build/src/lib.rs
@@ -319,3 +319,43 @@ fn env_os(key: impl AsRef<OsStr>) -> Result<OsString> {
     let key = key.as_ref();
     env::var_os(key).ok_or_else(|| Error::NoEnv(key.to_owned()))
 }
+
+
+#[test]
+fn test_include_dir_correct0() {
+    verify_include_dir(&PathBuf::from("some_crate/path/to/dir"), Path::new("some_crate"));
+}
+
+#[test]
+fn test_include_dir_correct1() {
+    verify_include_dir(&PathBuf::from("some_crate/path/to/dir"), Path::new("some_crate/path"));
+}
+
+#[test]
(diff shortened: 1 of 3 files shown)


gen/build/src/lib.rs
```diff
@@ -298,7 +302,61 @@ fn generate_bridge(prj: &Project, build: &mut Build, rust_source_file: &Path) ->
     Ok(())
 }
 
+// Ensures that the argument of the include!() macro is either starting with the crate name, or an absolute path
```
dtolnay, line 305: I think this check is somewhat too strict. For example the caller's build script is free to make any other directories available here and that wouldn't be a problem. Using https://crates.io/crates/pkg-config this could be a common pattern:

```rust
let python3 = pkg_config::probe_library("python3")?;
cxx_build::bridge("src/bridge.rs")
    .includes(python3.include_paths)
    .compile("repro");
```

```rust
extern "C" {
    include("Python.h");
    ...
}
```

dtolnay (changes requested):

gen/build/src/lib.rs
Bromeon: Under that circumstance, would it even be possible to check extra include directories inside cxx?  
We give the `cc::Build` object to the user and effectively lose control of how it's further configured.

dtolnay (commented):
I ended up landing a more specific check for includes relative to `.` or `..`, since we know those are meaningless in the Cargo-based workflow because the positioning of the generated code inside of Cargo's OUT_DIR is not specified.

Bromeon:
Okay.

Do you think the aggregation code I added in syntax/types.rs to collect all `include!` paths could still be used in the future (after I correct the set type)? As someone unfamiliar with the codebase, it took me a while to find the place to extract those. But we can of course add it only once it's really needed.

dtolnay:
Probably not. We do almost nothing with includes in this crate, and all of it is only iterating over them (to insert them into the output code). Iterating is easily done on only the syntax tree -- it doesn't require a more structured representation than that. The sets/maps in syntax/types.rs are for things where the generator needs to do more complicated lookup of type information.

https://github.com/dtolnay/cxx/blob/2cc2e3a4d13011e338b1624a6ced35f39f25894d/gen/src/write.rs#L26-L32

Bromeon:
Makes sense, thanks for the explanation! 🙂

PR #430, opened 2020-11-06
Fix clippy warnings

robinmoussu:


tests/test.rs
```diff
@@ -43,9 +43,13 @@ fn test_c_return() {
...
-    assert_eq!(
-        200.5_f64,
-        ffi::c_return_unique_ptr_vector_f64().into_iter().sum(),
+    assert!(
+        (200.5_f64
+            - ffi::c_return_unique_ptr_vector_f64()
+                .into_iter()
+                .sum::<f64>())
+        .abs()
+            <= f64::EPSILON
```
robinmoussu, line 52: I really don't think it's the best way to write it. For reason that I don't understand, it's not possible to write

```rust
#[allow(clippy::float_cmp)]
assert_eq!(
    200.5_f64,
    ffi::c_return_unique_ptr_vector_f64().into_iter().sum(),
```
I took a look at the implementation of `c_return_unique_ptr_vector_f64`, and I not 100% confident that all numbers can be represented with an exact f64, which would lead to an exact computation:

```
std::unique_ptr<std::vector<double>> c_return_unique_ptr_vector_f64() {
  auto vec = std::unique_ptr<std::vector<double>>(new std::vector<double>());
  vec->push_back(86.0);
  vec->push_back(75.0);
  vec->push_back(30.0);
  vec->push_back(9.5);
  return vec;
}
```

dtolnay: This computation is exact in standard IEEE floating point.

tests/test.rs
```diff
@@ -59,29 +63,29 @@ fn test_c_return() {
     assert_eq!(2021, ffi::c_return_sum(2020, 1));
     match ffi::c_return_enum(0) {
         enm @ ffi::Enum::AVal => assert_eq!(0, enm.repr),
-        _ => assert!(false),
+        _ => unreachable!(),
```
dtolnay, line 66: I think I would lean toward keeping this. The test is saying that reaching this codepath constitutes an assertion failure, not that reaching it is impossible.

tests/test.rs
```diff
@@ -59,29 +63,29 @@ fn test_c_return() {
...
         enm @ ffi::ABEnum::ABAVal => assert_eq!(0, enm.repr),
-        _ => assert!(false),
+        _ => unreachable!(),
     }
 }
 
 #[test]
 fn test_c_try_return() {
-    assert_eq!((), ffi::c_try_return_void().unwrap());
+    assert!(ffi::c_try_return_void().is_ok());
```
dtolnay, line 88: I think this makes it a worse test -- previously it was also testing that the T type of the Result in the macro-generated signature is correctly () as intended, not just that the result is Ok.

dtolnay (commented):
Thanks for the PR, though I disagree with all of these changes and would prefer not to assume we should do whatever clippy wants. I opened #433 to add `allow` for these lints instead.

tests/test.rs
robinmoussu: I tried to print the hexadecimal representation of the number, but without success. The trait isn't implemented for `f64`. That's good to know.

robinmoussu:
> Thanks for the PR, though I disagree with all of these changes and would prefer not to assume we should do whatever clippy wants. I opened #433 to add allow for these lints instead.

After re-reading what I did, I agree 100% with your move. Those suggestions where effectively bad.

tests/test.rs
dtolnay: It's just integer_mantissa \* 2\^exponent. e.g. 86\*2\^0 + 75\*2\^0 + 30\*2\^0 + 19\*2\^-1 = 401\*2\^-1. You get loss of precision for things like 0.1 which is approximated by 3602879701896397\*2\^-55.

robinmoussu: That's right, I thought that the issue was between '1.0 + 2.0 != 3.0`, but it's between `0.1 + 0.2 != 0.3`.

Issue #248, opened 2020-07-31
Return Result from rust functions?


lahwran:
Hey, what types can I return from a rust function as the second parameter of Result?

dtolnay:
- For Rust functions called from C++: **anything with a Display impl**. You'll need to declare the function signature inside the `mod ffi` as returning `-> Result<T>` without an explicit error type, but the actual implementation outside of the `mod ffi` can have any error type with Display. Returning an error will turn into an exception in C++ with type `rust::Error` whose message is the Display of your Rust error.

- For C++ functions called from Rust: if you declare the return type `-> Result<T>` then the error type you get when calling is always `cxx::Exception`. The Display impl of Exception will give you the message of the exception thrown by C++ according to std::exception::what (https://en.cppreference.com/w/cpp/error/exception/what).

PR #432, opened 2020-11-06
Documents how Result and exception can be used

robinmoussu:
Solve #248


dtolnay (commented):
Thanks for the PR! I am closing because a readme is not intended for exhaustive documentation at this level of detail. I am making exhaustive documentation in #179 which already includes this material.

robinmoussu:
That's perfect. I often create issue when people says that X should be in doc, but if you already added it, it's even better. I was also surprised that I didn't saw a mention of a book in the Readme but didn't realized it could just be currently being written!

PR #449, opened 2020-11-11
Switch intro example to blobstore client

dtolnay:
The new example is better aligned with real world usage.

This is also going to serve as the landing page example on https://cxx.rs.


dtolnay:
Rendered: https://github.com/dtolnay/cxx/tree/4ca366fa57662f0753ddb3400ebba592fcba70af#example

PR #456, opened 2020-11-11
Enforce unsafe surrounding block on safe-to-call functions

dtolnay:
As described in https://www.reddit.com/r/rust/comments/ielvxu/the_cxx_debate/g2i1r7a/.


dtolnay (commented):
To be landed next week with the 1.48 rustc release.

Issue #458, opened 2020-11-12
Support array in struct?


XiangpengHao:
What's your opinion on supporting array in struct? @dtolnay 
I assume this should be straightforward:

```rust
#[cxx::bridge]
mod ffi {
    struct Foo {
        a: [i32; 16],
        b: [Vec<usize>; 32],
    }
}
```

And it should be translated to 
```cpp
...
struct Foo;
struct Foo final {
  int32_t a[16];
  ::rust::Vec<size_t> b[32];
};
```

I'm not sure this is 100% safe as I'm new to both languages. If you're open to support this, I'll send a draft pull request.

dtolnay:
I am open to supporting this.

`std::array` would probably be a friendlier equivalence.

```cpp
struct Foo final {
  std::array<int32_t, 16> a;
  std::array<rust::Vec<size_t>, 32> b;
};
```

francesco-cattoglio:
Support for std::array would be awesome, I think it is one of the most useful stl bits imo.

dtolnay:
Implemented in #503, released in cxx 1.0.4!

PR #459, opened 2020-11-12
Allow array inside shared struct definition

XiangpengHao:
Address #458 

It translates 
```rust
#[cxx::bridge]
mod ffi {
    struct Foo {
        a: [i32; 0xf],
        b: [Vec<usize>; 3_2],
    }
}
```
to
```cpp
struct Foo final {
  std::array<int32_t, 15> a;
  std::array<rust::Vec<size_t>, 32> b;
};
```
The current implementation requires:
1. The size field must be an integer literal
2. The literal should be able to parse to usize
3. The type needs to be one of primitive type, shared struct, RustBox, RustVec, CxxVector, UniquePtr

I'm not sure if we can simply replace `[usize; 16]` to `std::array<uint64_t, 16>`.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -172,6 +172,9 @@ fn pick_includes_and_builtins(out: &mut OutFile, apis: &[Api]) {
                 out.include.cstdint = true;
                 out.builtin.rust_slice = true;
             }
+            Type::Array(_) => {
+                out.include.array = true;
+            }
             Type::Ref(_) | Type::Void(_) => {}
         }
     }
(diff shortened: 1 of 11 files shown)


dtolnay (commented):
Thank you, this is a good start. Some of the checks seem off though -- I am following up in #503.

For example:

```rust
use cxx::CxxString;

#[cxx::bridge]
mod ffi {
    extern "Rust" {
        fn p(array: &[CxxString; 3]);
    }
}

fn p(array: &[CxxString; 3]) {
    println!("{:?}", array);
}
```

```cpp
#include "test/src/bridge.rs.h"

int main() {
  std::array<std::string, 3> array = {"a", "b", "c"};
  p(array);
}
```

Under this PR it compiles and prints the following, which is wrong:

```console
["a", "a", "a"]
```

PR #524, opened 2020-11-28
boost::scoped_ptr demo

dtolnay:
Example code in response to https://github.com/dtolnay/cxx/issues/251#issuecomment-735275024.

Note this line is using `cxx::kind::Opaque`:

https://github.com/dtolnay/cxx/blob/bb4fda0682611b11c494abee2ad752114e88058b/scoped_ptr_demo/src/main.rs#L16

because passing a boost::scoped_ptr by value across the FFI is not possible due to both of the following being true:

1. It uses the "small class" ABI in C++ which Rust has no representation for. Any Rust data structure will always either look like "small POD" or "big classes" as far as the ABI in C++.

2. It exposes no way to exit out of "small class" state (analogous to `unique_ptr<T>::release` which is what we use for making UniquePtr work by value on the language boundary).

For a different smart pointer for which one or both of these doesn't apply you'd probably use `cxx::kind::Trivial` because Rust's move semantics are non-problematic for anything pointer-like.

### DIFF
--- a/scoped_ptr_demo/Cargo.toml
+++ b/scoped_ptr_demo/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "scoped-ptr-demo"
+version = "0.0.0"
+edition = "2018"
+links = "scoped-ptr-demo"
+publish = false
+
+[dependencies]
+cxx = "1.0"
+
+[build-dependencies]
+cxx-build = "1.0"
+
+[workspace]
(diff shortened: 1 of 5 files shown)


dtolnay:
#525 shows an approach where item 1 in https://github.com/dtolnay/cxx/pull/524#issue-529004118 does not apply -- it just works around "small class" by adding padding, which isn't great, but is compatible with Rust.

Issue #487, opened 2020-11-19
Opaque Rust type in shared_ptr: invalid application of ‘sizeof’ to incomplete type


venslu:
I want to pass the referenced type from Rust to C++, but the error message is as follows when compiling. Why? How can I pass  TransactionClient(struct type, defined in Rust) to C++ body? Please give me an example to solved it.

**error: invalid application of ‘sizeof’ to incomplete type ‘bgraph::storage::TransactionClient’**
```
  running: "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-I" "/home/work/tikv-rust/rust/test3/tikv-handler/target/debug/build/tikv-handler-a1bd85c7fdbe0d5e/out/cxxbridge/include" "-I" "/home/work/tikv-rust/rust/test3/tikv-handler/target/debug/build/tikv-handler-a1bd85c7fdbe0d5e/out/cxxbridge/crate" "-Wall" "-Wextra" "-std=c++14" "-o" "/home/work/tikv-rust/rust/test3/tikv-handler/target/debug/build/tikv-handler-a1bd85c7fdbe0d5e/out/src/tikv_handler.o" "-c" "src/tikv_handler.cpp"
  cargo:warning=In file included from /usr/include/c++/7/bits/shared_ptr_base.h:58:0,
  cargo:warning=                 from /usr/include/c++/7/bits/shared_ptr.h:52,
  cargo:warning=                 from /usr/include/c++/7/memory:81,
  cargo:warning=                 from /home/work/tikv-rust/rust/test3/tikv-handler/target/debug/build/tikv-handler-a1bd85c7fdbe0d5e/out/cxxbridge/crate/tikv-handler/include/tikv_handler.h:9,
  cargo:warning=                 from src/tikv_handler.cpp:6:
  cargo:warning=/usr/include/c++/7/ext/aligned_buffer.h: In instantiation of ‘struct __gnu_cxx::__aligned_buffer<bgraph::storage::TransactionClient>’:
  cargo:warning=/usr/include/c++/7/bits/shared_ptr_base.h:514:35:   required from ‘class std::_Sp_counted_ptr_inplace<bgraph::storage::TransactionClient, std::allocator<bgraph::storage::TransactionClient>, (__gnu_cxx::_Lock_policy)2>::_Impl’
[...]

PR #526, opened 2020-11-29
Issue 487 minimal repro

dtolnay:
#487 -- @venslu here's an example of stripping out unrelated code, like the dependency on tikv-client, async stuff, CxxVector stuff, etc. It reproduces the same error message but is a lot easier for someone else to help with.


venslu:
I understand，thank you very much

PR #530, opened 2020-11-30
Add Str(const String &) noexcept constructor

dtolnay:


cgwalters:
I was actually working on a PR to do this a day or two ago, thanks for this!

Issue #150, opened 2020-04-26
Improve C++ iterator functionality of rust::Vec<T>


dtolnay:
We currently give rust::Vec an extremely basic const_iterator, but it definitely doesn't support all the operations it should. I'm guessing https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator is the right list to follow.

Additionally we'll want a non-const iterator as well.

dtolnay:
Vec non-const iterator added in #509.

Random access iterator methods added in #538.

PR #538, opened 2020-12-03
Implement the random access iterator for rust::Vec

mindv0rtex:
My attempt at addressing #150.


dtolnay (approved):
Thanks!

On the topic of `std::sort` --- we are also going to want a `std::swap` overload for extern "Rust" opaque types. C++ has no way otherwise to know the size of the Rust types.

PR #576, opened 2020-12-13
Document and static assert assumptions on size_t layout

dtolnay:
Rust specifies that usize is ABI compatible with C's uintptr_t: https://rust-lang.github.io/unsafe-code-guidelines/layout/scalars.html#isize-and-usize. However there is no direct Rust equivalent for size_t. C does not guarantee that size_t and uintptr_t are compatible. In practice though, on all platforms supported by Rust, they are identical for ABI purposes. See the libc crate which unconditionally defines libc::size_t = usize. We expect the same here and the assertions in this PR are just to explicitly document that.

**Note that no assumption is made about C++ name mangling of signatures containing these types, not here nor anywhere in CXX.**

### DIFF
--- a/src/cxx.cc
+++ b/src/cxx.cc
@@ -249,6 +249,23 @@ std::ostream &operator<<(std::ostream &os, const Str &s) {
   return os;
 }
 
+// Rust specifies that usize is ABI compatible with C's uintptr_t.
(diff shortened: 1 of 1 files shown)


dtolnay:
Some background on this topic:

- https://github.com/rust-lang/rust-bindgen/issues/1671
- https://github.com/rust-lang/rust-bindgen/issues/1901
- https://github.com/eqrion/cbindgen/issues/239

Issue #224, opened 2020-06-22
rust::Str and rust::String not being null terminated is not obvious


TheZoq2:
First of all, thanks for this great library, with very little effort, it allows me to write a nice rust frontend fro my c++ project.

When I started using it to pass some strings from the rust side, I quickly found the Builtin types table, and given that a `data` function is available on `rust::Str`, I assumed that it would be fine to copy the content using `std::string(rust_string.data())`. Obviously this turned out to not be the case, but I don't think that's clear, at least from the parts of the documentation that I have read.

Would it be a good idea to put an entry in the "restrictions" column saying that this is the case?

dtolnay:
The intended reference for the API of the C++ types is the header: https://github.com/dtolnay/cxx/blob/0.3.4/include/cxx.h (for now; at least until #179 is closed). This is probably where you found the `data` function.

The header has a comment on Str::data and String::data indicating that there is no null terminator:

https://github.com/dtolnay/cxx/blob/bbd2620921e921825575810a9fd40712c378d6aa/include/cxx.h#L37-L38

https://github.com/dtolnay/cxx/blob/bbd2620921e921825575810a9fd40712c378d6aa/include/cxx.h#L66-L67

TheZoq2:
> This is probably where you found the data function.
Not quite, I did find it when noticing a problem with the strings, but I initially found it using my editor autocomplete. I guess this issue is mostly a "I read the bare minimum amount of docs and got surprised when things went wrong :)"

xfix:
`rust::Str` is like `std::string_view` which is not guaranteed to be null terminated.

cgwalters:
I'm actually converting a C codebase to "compile in C++ mode" purely so I can use this project.  And starting from a world wher e NUL terminated C strings are pervasive then adding C++ `std::string` and Rust `String` is very painful so far (basically need to do `auto s = std::string(rustobject->getstring()); some_c_function(s.c_str())`.

In some cases on the Rust side we've been using https://crates.io/crates/c_utf8 (which...IMO should probably be part of Rust std but that's a whole other discussion).  Using that one can avoid a copy.  Perhaps this crate could have a feature gate to enable adding that as a binding?

dtolnay:
How about adding a c_str() member function for rust::String? #586

PR #586, opened 2020-12-18
Add c_str() member function for rust::String

dtolnay:
Matches https://en.cppreference.com/w/cpp/string/basic_string/c_str.
#224

### DIFF
--- a/book/src/binding/string.md
+++ b/book/src/binding/string.md
@@ -33,6 +33,8 @@ public:
   size_t size() const noexcept;
   size_t length() const noexcept;
 
+  const char *c_str() noexcept;
+
   using iterator = char *;
   iterator begin() noexcept;
   iterator end() noexcept;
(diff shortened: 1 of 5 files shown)


cgwalters (approved):
Oh neat, this is an elegant solution.  The only thing that makes me slightly uncomfortable about this is conceptually we may end up mutating data that might be owned by the Rust side under a regular non-`mut` reference, if we returned an `&String` right?  But OTOH this is really a sort of "interior mutability" that should be invisible to anything accessing Rust side.  So it's safe from that PoV.

dtolnay:
You can't call this on a &amp;String because the member function is non-const.

cgwalters:
> You can't call this on a &String because the member function is non-const.

OK right.  This will help in a few cases I have now at least, though I think the case of having C++ iterate over data owned by Rust is going to be very common in at least my code, but I am going to try out using CUtf8 as an extern type.

bridiver:
is there a reason why `rust::Str` doesn't also have `c_str`? It would be much easier to work with in cpp without copying if it returned a null terminated string like `rust::String::c_str`

PR #613, opened 2020-12-29
add support for .hh and .hpp headers

permosegaard:
just a minor change to allow headers with extentions other than `.h` to used

### DIFF
--- a/gen/build/src/lib.rs
+++ b/gen/build/src/lib.rs
@@ -425,8 +425,9 @@ fn best_effort_copy_headers(src: &Path, dst: &Path, max_depth: usize) {
             }
             Ok(file_type) if file_type.is_file() => {
                 let src = entry.path();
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

gen/build/src/lib.rs
```diff
@@ -425,8 +425,9 @@ fn best_effort_copy_headers(src: &Path, dst: &Path, max_depth: usize) {
             }
             Ok(file_type) if file_type.is_file() => {
                 let src = entry.path();
-                if src.extension() != Some(OsStr::new("h")) {
-                    continue;
+                match src.extension().map(|x| x.to_str()).flatten() {
```
dtolnay, line 428: FYI a more compact way to write this is:

```suggestion
                match src.extension().and_then(OsStr::to_str) {
```

PR #636, opened 2021-01-02
Remove rust::Slice repr conversion in C++

dtolnay:
These layout conversions between pod and non-pod were made obsolete by #387, as since then `rust::Str` and `rust::Slice<T>` are correctly implemented as pod.


dtolnay:
I think this'll work but one additional workaround is required by MSVC. https://stackoverflow.com/a/57429504/6086311

dtolnay:
Rebased on #637.

PR #641, opened 2021-01-02
Include a launch config for cargo tests

capickett:
First time using VS Code to develop and I found this config made it easier to jump into breakpoint debugging the cargo tests.

"Debug cargo tests" now shows up in the drop-down, and setting a breakpoint in CXX or rust correctly stops.

Debugger in VSCode

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,6 +12,18 @@
                     "kind": "bin"
                 }
             }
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Great! I haven't been using VS Code but it's awesome that this works.

PR #642, opened 2021-01-02
Add swap members + ADL functions

### LINKED ISSUE #58: Provide a `swap` implementation for types in cxxbridge.h
As recommended in #8. String, Str, and Box should have an efficient `swap`.

capickett:
Fixes #58.

Adds `swap` members to each of `Slice`, `Str`, `Vec`, `Box`, and `String`. Also adds private friend functions so that Argument Dependent Lookup may function via the `using std::swap;` idiom.

### DIFF
--- a/book/src/binding/box.md
+++ b/book/src/binding/box.md
@@ -25,6 +25,8 @@ public:
   explicit Box(const T &);
   explicit Box(T &&);
 
+  void swap(Box &) noexcept;
+
   Box &operator=(const Box &);
   Box &operator=(Box &&) noexcept;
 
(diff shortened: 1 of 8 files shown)


capickett:
One comment @dtolnay - I noticed some members were implemented later in the header, whilst others were implemented in the cpp file. I wasn't clear on the rationale between when to implement in one or the other, so I tried to be consistent with existing code.

dtolnay:
> some members were implemented later in the header, whilst others were implemented in the cpp file. I wasn't clear on the rationale

Function templates need to be in a header in C++ for the caller to instantiate. Plain non-template functions go in the source file.

dtolnay (approved):

capickett:
@dtolnay yep that makes a ton of sense! The holidays have gotten to me :)

PR #643, opened 2021-01-02
Remove Str and Slice swaps

dtolnay:
Both of these types are "is_trivially_move_constructible" and "is_trivially_move_assignable", so the default std::swap would already correctly do exactly the same thing that these are doing.

We keep the custom swap on Box, String, Vec which are not trivially movable, their moves do stuff to the source object to move around ownership.

FYI @capickett

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -106,8 +106,6 @@ public:
   std::size_t size() const noexcept;
   std::size_t length() const noexcept;
 
-  void swap(Str &) noexcept;
-
   // Important in order for System V ABI to pass in registers.
   Str(const Str &) noexcept = default;
   ~Str() noexcept = default;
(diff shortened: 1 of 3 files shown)


capickett:
A small note here is that this changes means that

```
rust::Str s1 = "abc";
rust::Str s2 = "def";
s1.swap(s2);
```

Will now no longer compile. Referencing the STL, it appears `std::string_view` exposes a member `swap` https://en.cppreference.com/w/cpp/string/basic_string_view/swap. Perhaps worth keeping the member and implementing as:

```
void Str::swap(Str &rhs) noexcept {
  std::swap(*this, rhs);
}
```

dtolnay:
Thanks! Good catch, I was only paying attention to the "std::swap"-style usage. Thanks for the link, it seems clear that we should have the member functions to match string_view. I've put those back in #646.

PR #666, opened 2021-01-09
WIP: Allow for Clone bound on extern Rust type + Box clone

### LINKED ISSUE #105: Wire up Box's copy constructor to call Clone on opaque Rust types
Currently C++ gets to use the Box\<T\> copy constructor only if T is a shared struct. If T is an opaque Rust type then copy constructing Box\<T\> will not compile.

We should provide a way to constrain that some opaque Rust type implements Clone and then expose the Clone impl to C++ via the Box\<T\> copy constructor.

```rust
mod ffi {
    extern "Rust" {
        type ClonableR: Clone;
    }

    extern "C" {
        fn f(b: Box<ClonableR>); // C++ will be able to clone b
    }
}
```

capickett:
Fixes #105.

First, extends the parsing syntax to allow for `Clone` bounds added to an extern Rust type.
Then, introduces a public `Box clone() const noexcept;` member on `Box<T>` that will only be callable if `T` is declared as `Cloneable`.
Finally, exposes a clone FFI function callable from C++ to clone the Box. These bindings are only generated if the type derives `Clone` or has a trait bound for `Clone`.

Tested with `cargo test`

Still todo:

- [ ] Fix trybuilt test cases
- [ ] Make sure the approach makes sense.
- [ ] Extend documentation to describe the new behavior.

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -30,6 +30,54 @@ template <typename T>
 class impl;
 }
 
+#ifndef CXXBRIDGE1_TRAITS
+#define CXXBRIDGE1_TRAITS
+namespace detail {
+template <typename... Ts>
+struct make_void {
+  using type = void;
+};
+
+template <typename... Ts>
+using void_t = typename make_void<Ts...>::type;
+
+template <typename Void, template <typename...> class, typename...>
+struct detect : std::false_type {};
+template <template <typename...> class T, typename... A>
+struct detect<void_t<T<A...>>, T, A...> : std::true_type {};
+
+template <template <typename...> class T, typename... A>
+using is_detected = detect<void, T, A...>;
+
+template <typename T>
+using detect_IsRelocatable = typename T::IsRelocatable;
+template <typename T>
+struct get_IsRelocatable
+    : std::is_same<typename T::IsRelocatable, std::true_type> {};
+
+template <typename T>
+using detect_IsCloneable = typename T::IsCloneable;
+template <typename T>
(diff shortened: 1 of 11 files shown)


capickett:
Ominous #666

include/cxx.h
```diff
@@ -30,6 +30,54 @@ template <typename T>
...
+
+template <typename T>
+using detect_IsCloneable = typename T::IsCloneable;
+template <typename T>
+struct get_IsCloneable : std::is_same<typename T::IsCloneable, std::true_type> {
+};
+} // namespace detail
+
+template <typename T>
+struct IsRelocatable
```
capickett, line 66: I wanted to share some of the detection bits here with `IsRelocatable` to make `Box::clone` conditionally compiled. I collapsed these into a single "traits" section, but if you'd prefer, I can keep "IsCloneable" separate

syntax/resolve.rs
```diff
@@ -1,10 +1,12 @@
-use crate::syntax::{Lifetimes, NamedType, Pair, Types};
+use crate::syntax::{Derive, Lifetimes, NamedType, Pair, Types};
 use proc_macro2::Ident;
 
 #[derive(Copy, Clone)]
 pub struct Resolution<'a> {
```
capickett, line 5: This change I was also uncertain of. I'm unsure the intended scope of `Resolution`, but it seemed to me that bounds and derives were as associated to a type as `generics`. Happy to split this out to a separate collection in `Types`

capickett:
Don't have time to pick this back up

PR #677, opened 2021-01-15
Improve error message when using self by-value

### LINKED ISSUE #577: Improve error message on signatures containing self by value
The current error is shown as:

```console
error[cxxbridge]: unsupported signature
  ┌─ /dev/stdin:1:54
  │
1 │ #[cxx::bridge] mod ffi { extern "C++" { type T; fn f(self); }}
  │                                                      ^^^^ unsupported signature
```

Instead it should say what to use instead. `&self` or `&mut self` for shared structs and opaque Rust types, `self: Pin<&mut Self>` for an opaque C++ type, `Box<T>` or `UniquePtr<T>` if ownership needs to be transferred depending on opaque Rust or opaque C++ type.

fanzeyi:
Fixes #577 .

----

Hi David, I'm picking up some help-wanted issue from the list and this one seems to be a good start. Let me know if there is anything else I can help. (There are a few error message improvement issues I'm going to look into for now :D).

### DIFF
--- a/syntax/parse.rs
+++ b/syntax/parse.rs
@@ -567,7 +567,21 @@ fn parse_extern_fn(
                     });
                     continue;
                 }
-                return Err(Error::new_spanned(arg, "unsupported signature"));
+
+                match lang {
+                    Lang::Rust => {
+                        return Err(Error::new_spanned(
+                            arg,
+                            "cannot taking self by-value, consider using `&self` or `&mut self` instead"
+                        ));
+                    }
+                    Lang::Cxx => {
+                        return Err(Error::new_spanned(
+                            arg,
+                            "cannot taking self by-value, consider using `self: Pin<&mut Self>`, `self: Box<T>` or `self: UniquePtr<T>` instead"
+                        ));
+                    }
+                }
             }
             FnArg::Typed(arg) => {
                 let ident = match arg.pat.as_ref() {
--- a/tests/ui/self_suggestion.rs
+++ b/tests/ui/self_suggestion.rs
@@ -1,0 +1,12 @@
+#[cxx::bridge]
+mod ffi {
+    extern "Rust" {
+        fn self_rust(self);
+    }
+
+    extern "C++" {
+        fn self_c(self);
+    }
+}
+
+fn main() {}
--- a/tests/ui/self_suggestion.stderr
+++ b/tests/ui/self_suggestion.stderr
@@ -1,0 +1,11 @@
+error: cannot taking self by-value, consider using `&self` or `&mut self` instead
+ --> $DIR/self_suggestion.rs:4:22
+  |
+4 |         fn self_rust(self);
+  |                      ^^^^
+
+error: cannot taking self by-value, consider using `self: Pin<&mut Self>`, `self: Box<T>` or `self: UniquePtr<T>` instead
+ --> $DIR/self_suggestion.rs:8:19
+  |
+8 |         fn self_c(self);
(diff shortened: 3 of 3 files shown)


syntax/parse.rs
```diff
@@ -567,7 +567,21 @@ fn parse_extern_fn(
                     });
                     continue;
                 }
-                return Err(Error::new_spanned(arg, "unsupported signature"));
+
+                match lang {
+                    Lang::Rust => {
```
dtolnay, line 572: This is branching on the wrong thing -- the requirement for Pin i.e. the use of `&mut self` vs `self: Pin<&mut Self>` is not based on the source language of the function implementation, but on whether the receiver type is or is not opaque to Rust. Consider for example:

```rust
use core::pin::Pin;

#[cxx::bridge]
pub mod ffi {
    extern "C++" {
        type C;
    }

    extern "Rust" {
        type R;
        fn f(self: &mut R);
        fn f(self: Pin<&mut C>);  // Rust method on an opaque C++ type
    }
}

pub struct R;

impl R {
    fn f(&mut self) {}
}

impl ffi::C {
    fn f(self: Pin<&mut Self>) {}
}
```

where the C struct is:

```cpp
class C {
  void f() noexcept;
};
```

syntax/parse.rs
```diff
@@ -567,7 +567,21 @@ fn parse_extern_fn(
...
+                        return Err(Error::new_spanned(
+                            arg,
+                            "cannot taking self by-value, consider using `&self` or `&mut self` instead"
+                        ));
+                    }
+                    Lang::Cxx => {
+                        return Err(Error::new_spanned(
+                            arg,
+                            "cannot taking self by-value, consider using `self: Pin<&mut Self>`, `self: Box<T>` or `self: UniquePtr<T>` instead"
```
dtolnay, line 581: `self: UniquePtr<T>` is not allowed on stable Rust so this shouldn't suggest it. See https://github.com/rust-lang/rust/issues/44874.

```console
error[E0658]: `UniquePtr<T>` cannot be used as the type of `self` without the `arbitrary_self_types` feature
  --> src/main.rs:11:16
   |
11 |     fn f(self: UniquePtr<T>) {}
   |                ^^^^^^^^^^^^
   |
   = note: see issue #44874 <https://github.com/rust-lang/rust/issues/44874> for more information
   = help: add `#![feature(arbitrary_self_types)]` to the crate attributes to enable
   = help: consider changing to `self`, `&self`, `&mut self`, `self: Box<Self>`, `self: Rc<Self>`, `self: Arc<Self>`, or `self: Pin<P>` (where P is one of the previous types except `Self`)
```

Separately, `self: Box<T>` is not a thing that C++ is able to do, because it can only call non-static member functions of `T` on `T&`, `T const&`, `T&&`, and `T const&&`.

The way they'd want to transfer ownership is by having a static member function or non-member function where `Box<T>` or `UniquePtr<T>` is the argument type (as opposed to receiver).

tests/ui/self_suggestion.rs
```diff
@@ -0,0 +1,12 @@
+#[cxx::bridge]
+mod ffi {
+    extern "Rust" {
+        fn self_rust(self);
```
dtolnay, line 4: Consider also testing `self: R`.

dtolnay (changes requested):
Thanks!

dtolnay (commented):
Closing as inactive. Feel free to open a new PR whenever you get back to addressing the feedback!

Issue #561, opened 2020-12-11
Support moving into an rvalue reference argument


dtolnay:
Consider this code:

```rust
// src/lib.rs

#[cxx::bridge]
mod ffi {
    unsafe extern "C++" {
        include!("example/include/example.h");
        type Thing;
        fn f(thing: UniquePtr<Thing>);
    }
}
```

```cpp
#pragma once
#include <memory>

struct Thing {};

void f(std::unique_ptr<Thing> &&);
```

This would currently fail to build, with:

```console
example-760540e262f31cef/out/cxxbridge/sources/example/src/main.rs.cc: In function ‘void cxxbridge1$f(Thing*)’:
example-760540e262f31cef/out/cxxbridge/sources/example/src/main.rs.cc:34:46: error: invalid conversion from ‘void (*)(std::unique_ptr<Thing>&&)’ to ‘void (*)(std::unique_ptr<Thing>)’ [-fpermissive]
   34 |   void (*f$)(::std::unique_ptr<::Thing>) = ::f;
      |                                            ~~^
      |                                              |
      |                                              void (*)(std::unique_ptr<Thing>&&
```

because the cxx C++ code generator is expecting a signature `void f(std::unique_ptr<Thing>)` whereas the actual signature is `void f(std::unique_ptr<Thing> &&)`.

It would be good to support this better and interoperate between a Rust by-value argument and a C++ rvalue reference argument.

Maybe like:

```rust
extern "C++" {
    fn f(thing: #[move] UniquePtr<Thing>);
}
```

or this bastardization of C++ rvalue syntax:

```rust
extern "C++" {
    fn f(&&thing: UniquePtr<Thing>);
}
```

PR #679, opened 2021-01-15
RFC: Support rvalue argument

fanzeyi:
See #561.

I opted in the `foo: &&T` syntax instead so editors can still somewhat treat it as Rust code. `&&foo: T` is .. weird and out-of-place and `#[move]` will just break. Please let me know if there is a case for these two instead of `foo: &&T`.

This is a draft since I think relaxing the reference to reference check may have broken some other stuff which I haven't gotten time to test (e.g. what would be generated if I put && in a Rust function signature). 

The handling of `Type::Ref` is also a little manual, and maybe adding a field to `Ref` type to indicate rvalue reference can improve this.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -778,6 +778,12 @@ fn write_cxx_function_shim<'a>(out: &mut OutFile<'a>, efn: &'a ExternFn) {
         } else if out.types.needs_indirect_abi(&arg.ty) {
             out.include.utility = true;
             write!(out, "::std::move(*{})", arg.name.cxx);
+        } else if let Type::Ref(ref inner) = arg.ty {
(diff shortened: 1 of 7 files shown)


dtolnay (commented):
I would strongly prefer not to use `var: &&Type`. As a type, `&&Type` already has a well-defined widely used meaning in Rust that is different from this. It would be wrong to take a signature declared with `&&Type` as the argument type and make it into a function that requires the caller to pass `Type` by value.

Do you think you could brainstorm some alternative implementation options for this?

dtolnay (commented):
I'll go ahead and close since this has been sitting for a while, but please feel free to open a new PR whenever you get back to working on this. Thanks anyway!

PR #688, opened 2021-01-20
Clarify error for using enum as method receiver

### LINKED ISSUE #578: Improve error message on enum used as method receiver
```console
error[cxxbridge]: unrecognized receiver type
  ┌─ /dev/stdin:1:66
  │
1 │ #[cxx::bridge] mod ffi { enum Enum {} extern "Rust" { fn f(self: &Enum); }}
  │                                                                  ^^^^^ unrecognized receiver type
```

This needs to give some indication that member functions are not allowed by C++ on enums.

fletcherw:
Update the syntax checker to print a more specific error when an enum is
used as a method receiver, rather than printing a misleading "unrecognized
receiver type" error.

This is done because C++ prohibits enums from having member functions.

Fixes #578

### DIFF
--- a/syntax/check.rs
+++ b/syntax/check.rs
@@ -389,6 +389,8 @@ fn check_api_fn(cx: &mut Check, efn: &ExternFn) {
                 mutability = mutability,
             );
             cx.error(span, msg);
+        } else if cx.types.enums.contains_key(&receiver.ty.rust) {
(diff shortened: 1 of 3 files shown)


cgwalters (approved):
(Not really a cxx reviewer but) this looks sane to me.

dtolnay (approved):
Thanks!

PR #691, opened 2021-01-27
Updated Bazel Rust rules version

UebelAndre:
As of https://github.com/bazelbuild/rules_rust/commit/1fe23158e5316c17b2ee2a252ee7165c5d83cc93 the Bazel rust rules have been renamed from `io_bazel_rules_rust` to `rules_rust`. This PR updates the name here.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -2,39 +2,18 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("//tools/bazel:vendor.bzl", "vendor")
 
 http_archive(
-    name = "io_bazel_rules_rust",
-    sha256 = "5ed804fcd10a506a5b8e9e59bc6b3b7f43bc30c87ce4670e6f78df43604894fd",
-    strip_prefix = "rules_rust-fdf9655ba95616e0314b4e0ebab40bb0c5fe005c",
-    # Master branch as of 2020-07-30
-    url = "https://github.com/bazelbuild/rules_rust/archive/fdf9655ba95616e0314b4e0ebab40bb0c5fe005c.tar.gz",
-)
-
-http_archive(
-    name = "bazel_skylib",
-    sha256 = "97e70364e9249702246c0e9444bccdc4b847bed1eb03c5a3ece4f83dfe6abc44",
+    name = "rules_rust",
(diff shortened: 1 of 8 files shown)


tools/bazel/BUILD
```diff
@@ -0,0 +1,7 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
```
UebelAndre, line 1: For awareness, this is added to allow users to generate documentation through Bazel's [stardoc](https://github.com/bazelbuild/stardoc) tool. If a downstream user writes a `*.bzl` file that loads any of the `*.bzl` files here, they will need a [bzl_library](https://github.com/bazelbuild/bazel-skylib/blob/master/bzl_library.bzl) containing the files here available to their workspace in order to generate documentation.

UebelAndre:
@dtolnay Hey, I'm not sure what the PR process is but I'd greatly appreciate it if you could take a look at this when you get a moment 🙏 😅

WORKSPACE
```diff
@@ -2,39 +2,18 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
...
-    exec_triple = "x86_64-unknown-linux-gnu",
-    version = "1.48.0",
-)
-
-rust_repository_set(
-    name = "rust_1_48_darwin",
-    exec_triple = "x86_64-apple-darwin",
-    version = "1.48.0",
-)
+rust_repositories()
```
dtolnay, line 16: Could you please keep explicit control over what rustc version is used, even if it would currently be the default?

tools/bazel/vendor.bzl
```diff
@@ -48,7 +52,12 @@ def _log_cargo_vendor(repository_ctx, result):
         repository_ctx.execute(print, quiet = False)
 
 vendor = repository_rule(
-    attrs = {"lockfile": attr.label()},
+    doc = "A rule used to vender the dependencies of a crate in the current workspace",
```
dtolnay, line 55:
```suggestion
    doc = "A rule used to vendor the dependencies of a crate in the current workspace",
```

dtolnay (changes requested):
Thanks!

WORKSPACE
UebelAndre: Like so? Or did you want the platforms to be explicit too?

dtolnay (approved):
Thank you, looks good!

PR #692, opened 2021-01-27
Upgrade the dependency on codespan_reporting

MForster:
This is to avoid a downstream dependency to depend on multiple versions of that crate.


dtolnay (approved):
Thanks!

PR #699, opened 2021-01-31
Minor correction in the async example

wqfish:
The first argument of the closure is actually the context, not the sender. Rename it to `context` to make it clearer. Was slightly confused the first time I read it.


dtolnay (approved):
Thanks!

PR #700, opened 2021-02-01
Make rust_cxx_bridge usable from other repositories

hlopko:
Currently the `rust_cxx_bridge.bzl` has to be copied to the main repository (and edited) if it is to be used from a different repository than cxx's own one in Bazel. After giving the cxx repo a name and using the label with the repository name the bzl file is now usable from outside as well.

Thanks! :)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,3 +1,5 @@
+workspace(name = "cxx.rs")
+
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("//tools/bazel:vendor.bzl", "vendor")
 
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -1,3 +1,4 @@
+workspace(name = "cxx")
```
dtolnay, line 1: Is this workspace name expected to be globally unique in the downstream repo? `cxx` is sufficiently unique as the name of a Rust crate, but not necessarily as an identifier in a multi language codebase. Would it be idiomatic to use e.g. `cxx.rs`?

hlopko: I like `cxx.rs`, great idea! PTAL :)

dtolnay (approved):

PR #704, opened 2021-02-07
Re-export cc's parallel feature for cxx-build

edsrzf:
This allows building in parallel when the feature is enabled.

I wanted to try cxx-build with a fairly large code base and am hoping this will speed up the build. I may end up resorting to some other tool/configuration, but I can't see any reason not to try this first.


dtolnay (approved):
Thanks!

PR #717, opened 2021-02-15
Include license files in flags crate

khardix:
Other crates in this repo already include them, and it is a requirement (or at least a strong suggestion) for packaging it into Fedora.


dtolnay (approved):
Thanks!

Issue #592, opened 2020-12-23
char is not supported as function argument


alexeyr:
```
  error[cxxbridge]: unsupported type: char
     ┌─ tokenizers-cpp/pre_tokenizers.rs:56:52
     │
  56 │         fn char_delimiter_pre_tokenizer(delimiter: char) -> Box<PreTokenizer>;
     │                                                    ^^^^ unsupported type
```
Is the reason that some `u32` values are not valid `char`s? If so, this could be specified in the error message.

alexeyr:
It seems consistent with `rust::Str`/`String` to add the class `rust::Char` with `Char(char32_t)` constructor throwing exception on an invalid value and an implicit `operator char32_t()`.

david-cattermole:
Hello @alexeyr,

I don't think `char` is supported, but I think you'd want `std::os::raw::c_char` instead.
I think that will get translated into a char in C++ (although I haven't done it myself, yet).

I personally find looking at the cxx tests to be very helpful when looking for tested data type combinations.
https://github.com/dtolnay/cxx/blob/master/tests/ffi/lib.rs

alexeyr:
@david-cattermole I was interested in going the other way around, wrapping Rust code using `char` (and expecting it to translate to `char32_t`).

kraktus:
I would also be interested in having support for `char`

kraktus:
[cbdingen](https://github.com/eqrion/cbindgen/blob/master/docs.md) for instance support chars and use `char => uint32_t` conversion for the rust -> cpp side. For the cpp -> rust I think the simple fix of converting to char via `char::from_u32(uint32_t).expect("Valid char")` would enough for a first implementation.

PR #725, opened 2021-02-20
Add CxxChar to bind to C++ char32_t type

edsrzf:
This makes it possible to implement functions in Rust that accept a `char32_t`. The `CxxChar` type is necessary because C++ does not guarantee that `char32_t` contains a valid Unicode scalar value.

There is not yet a way for C++ code to create a Rust `char`.

I'll update the docs after getting feedback on the design and implementation.

Partly addresses #592.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -399,6 +399,8 @@ extern crate std;
 #[macro_use]
 mod macros;
 
+#[path = "cxx_char.rs"]
+mod char;
 mod exception;
 mod extern_type;
 mod function;
@@ -420,6 +422,7 @@ mod unwind;
 pub mod vector;
 mod weak_ptr;
 
+pub use crate::char::CxxChar;
 pub use crate::exception::Exception;
 pub use crate::extern_type::{kind, ExternType};
 pub use crate::shared_ptr::SharedPtr;
@@ -429,6 +432,13 @@ pub use crate::unique_ptr::UniquePtr;
 pub use crate::vector::CxxVector;
 pub use crate::weak_ptr::WeakPtr;
 pub use cxxbridge_macro::bridge;
+
+/// Synonym for `CxxChar`.
+///
+/// To avoid confusion with Rust's built-in char type you probably
+/// shouldn't import this type with `use`. Instead, write `cxx::Char`, or
+/// import and use `CxxChar`.
+pub type Char = CxxChar;
 
 /// Synonym for `CxxString`.
 ///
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -430,6 +433,13 @@ pub use crate::vector::CxxVector;
 pub use crate::weak_ptr::WeakPtr;
 pub use cxxbridge_macro::bridge;
 
+/// Synonym for `CxxChar`.
+///
+/// To avoid confusion with Rust's built-in char type you probably
+/// shouldn't import this type with `use`. Instead, write `cxx::Char`, or
+/// import and use `CxxChar`.
+pub type Char = CxxChar;
```
edsrzf, line 441: I wasn't entirely sure of the reason for these aliases, so just following the existing pattern. 🤷‍♂️

dtolnay (commented):
Thanks for the PR! This implementation looks good. However I wonder if it's necessary to have a repr(transparent) newtype wrapper, or if we could use just `u32` as the type on the Rust side? Do you have a use case where providing a wrapper type is important?

edsrzf:
The problem with using straight `u32` for the C++ declaration is that if we do this:

```rust
fn rust_fn(ch: u32) {}

#[cxx::bridge]
pub mod ffi {
    extern "Rust" {
        fn rust_fn(ch: u32);
    }
}
```

...then the generated C++ declaration will look something like this:

```c++
void rust_fn(::std::uint32_t ch) noexcept;
```

...and uint32_t and char32_t are distinct C++ types, mangled differently. ([Reference link](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-builtin) for the IA64 ABI followed by gcc and clang.)

It's not specifically important to me to provide a wrapper type, but it is important to me to be able to generate C++ function signatures compatible with `char32_t`. If there's a better way to accomplish this, then I'm all ears.

dtolnay (changes requested):
Okay in that case I would ask that there not be a newtype wrapper. `char32_t` in the bridge should use `u32` in Rust and `::std::char32_t` in C++. You can also provide a type alias for `u32` as something like `cxx::num::char32_t` for downstream code to use if they choose.

edsrzf:
So just to be clear, would you expect to be able to write:

```rust
fn rust_fn(ch: u32) {}

#[cxx::bridge]
pub mod ffi {
    extern "Rust" {
        fn rust_fn(ch: char32_t);
    }
}
```

...and then generate a C++ declaration using `char32_t`?

dtolnay:
Right. That is how the standard library and libc do C integer types.

https://doc.rust-lang.org/1.50.0/std/os/raw/index.html
https://docs.rs/libc/0.2.91/libc/#types

Canonically the rust_fn implementation would be written using cxx::num::char32_t though.

PR #727, opened 2021-02-23
Split ExternType out into a standalone crate

sbrocket:
As mentioned in trait/README.md, this is useful for implementing ExternType for types in a library crate without requiring all users of the library to link in the full cxx along with its C++ symbols.

----

Unfortunately the optional dependency approach (recommended in #726 which was closed) does not work with our non-Cargo build system. Our build system (GN/Ninja) has no equivalent. I could do something like it by splitting the library into 2 variants - once with the feature, once without - and requiring consumers to manually pick the variant they want, but I believe that won't work as soon as the dependency graph ends up including both versions (which is extremely likely in my use case, since I want to use this in a crate depended on by ~every Fuchsia Rust crate).

I believe Cargo works around this using name mangling, but I think even that approach has limitations in that objects from the different versions are incompatible? Or perhaps in this case Cargo would simply figure out whether the feature is needed for anything in the dependency graph. In any case, that approach is not available outside of Cargo.

Also, the approach you recommend is still available for environments where Cargo is in use. Perhaps we could recommend in the README that people prefer that approach unless it does not work for them?


dtolnay (commented):
Let me reopen the first PR. Will respond in a bit.

PR #743, opened 2021-03-18
Implement Eq, Ord, PartialOrd, Hash for CxxString

### LINKED ISSUE #737: Trait impls for CxxString:  Hash, Ord, PartialOrd, Eq
I need to use `&CxxString` as the key of a BTreeMap or HashMap. However there are some trait impls currently missing, which we should just make delegate to the underlying `&[u8]`. BTreeMap requires `Ord` and HashMap requires `Hash + Eq`.

b05902132:
Closes #737.

This is implemented by delegation to `&[u8]`, as suggested.

I also implemented `PartialOrd<str>`  for `CxxString` and `PartialOrd<CxxString>` for `str`, since `PartialEq<str>` is already implemented for `CxxString` and vice versa. I added tests for comparisons between `str` and `CxxString`, just in case.

### DIFF
--- a/src/cxx_string.rs
+++ b/src/cxx_string.rs
@@ -183,6 +183,38 @@ impl PartialEq<str> for CxxString {
     }
 }
 
+impl Eq for CxxString {}
+
+impl PartialOrd for CxxString {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        self.as_bytes().partial_cmp(other.as_bytes())
+    }
+}
+
+impl PartialOrd<str> for CxxString {
+    fn partial_cmp(&self, other: &str) -> Option<core::cmp::Ordering> {
+        self.as_bytes().partial_cmp(other.as_bytes())
+    }
+}
+
+impl PartialOrd<CxxString> for str {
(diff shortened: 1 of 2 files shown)


b05902132:
Oh by the way the standard library implements `Hash` for `str` as this. https://doc.rust-lang.org/src/core/hash/mod.rs.html#598-603. They add additional `0xff` when hashing.
Do we follow the standard library? If we don't, then we should reconsider implementing `PartialEq<str>` for `CxxStrins` and some other traits.

dtolnay (approved):
Thanks!

dtolnay:
> If we don't, then we should reconsider implementing `PartialEq<str>` for `CxxString` and some other traits.

Why? I don't know of any relationship that PartialEq would impose on the hash implementation.

b05902132:
> > If we don't, then we should reconsider implementing `PartialEq<str>` for `CxxString` and some other traits.
> 
> Why? I don't know of any relationship that PartialEq would impose on the hash implementation.

I thought if two objects are equal, then they should have same hash? I may be mistaken though.

dtolnay:
That is definitely not required.

PR #748, opened 2021-03-23
Detect raw pointer argument inside of function pointer

dtolnay:
Fixes https://github.com/dtolnay/cxx/pull/689#discussion_r599192164.

Follow-up to #689.


dtolnay:
I filed #749 to follow up on the code quality issue of having this check duplicated separately for extern functions vs function pointers.

PR #752, opened 2021-03-23
Disable lint jobs on pull requests

dtolnay:
We have been using this approach on the serde repo since last year and it works well (https://github.com/serde-rs/serde/blob/v1.0.125/.github/workflows/ci.yml#L155). It's quite common for Clippy to have low-value lints that we are better off disabling, but a contributor is not necessarily equipped to decide whether to suppress or resolve a lint, and consequently a clippy CI job can end up encouraging PRs submitted with needless workarounds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,7 @@ jobs:
   clippy:
     name: Clippy
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Screenshot from 2021-03-22 19-49-30

PR #779, opened 2021-03-26
Allow mutable return reference on signatures with any lifetime in argument

### LINKED ISSUE #765: Allow mutable return reference on signatures with any lifetime in argument
As of #608 now that extern types defined in the bridge can be generic over lifetimes, it no longer makes sense to expect specifically a reference in one of the function arguments of a function that returns a mutable reference. The following two signatures would be perfectly valid:

```rust
#[cxx::bridge]
mod ffi {
    extern "Rust" {
        type Thing<'a>;
    }

    unsafe extern "C++" {
        fn f<'a>(t: Box<Thing<'a>>) -> &'a mut i32;
        fn g(t: Box<Thing>) -> &mut i32;  // same thing, with lifetime elision
    }
}

pub struct Thing<'a>(&'a mut i32);
```

Current behavior:

```console
error[cxxbridge]: &mut return type is not allowed unless there is a &mut argument
  ┌─ src/main.rs:8:9
  │
8 │         fn f<'a>(t: Box<Thing<'a>>) -> &'a mut i32;
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ &mut return type is not allowed unless there is a &mut argument

error[cxxbridge]: &mut return type is not allowed unless there is a &mut argument
  ┌─ src/main.rs:9:9
  │
9 │         fn g(t: Box<Thing>) -> &mut i32;
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ &mut return type is not allowed unless there is a &mut argument
```

dtolnay:
Fixes #765.

### DIFF
--- a/syntax/check.rs
+++ b/syntax/check.rs
@@ -1,5 +1,6 @@
 use crate::syntax::atom::Atom::{self, *};
 use crate::syntax::report::Errors;
+use crate::syntax::visit::{self, Visit};
 use crate::syntax::{
     error, ident, trivial, Api, Array, Enum, ExternFn, ExternType, Impl, Lang, NamedType, Ptr,
     Receiver, Ref, Signature, SliceRef, Struct, Trait, Ty1, Type, TypeAlias, Types,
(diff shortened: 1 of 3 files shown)


jaques-sam:
Hello, I don't really get why an argument needs to be provided in that case.
How do I define an extern C++ function call like this one?:
```cpp
Logger &logger();
```

I tried this, but I still get this weird error:
```rust
     ┌─ src/main.rs:32:9
     │
  32 │         fn logger<'static>() -> Pin<&'static mut Logger>;  // not working :'(
     │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ &mut return type is not allowed unless there is a &mut argument
```

Issue #830, opened 2021-04-13
Duplicate const generated on cxx::bridge file in C++


andfoy:
Thanks for your work on this useful library! Right now I'm using `cxx::bridge` to provide Elixir bindings for PyTorch via Rustler (https://github.com/andfoy/extorch). I'm trying to use a  `rust::Slice` with `std::shared_ptr<CrossTensor>`, where `CrossTensor` is a C++ opaque type as follows:

```rust
/// Index a tensor using a list of tensors.
fn index(
    tensor: &SharedPtr<CrossTensor>,
    indices: &[&SharedPtr<CrossTensor>]
) -> Result<SharedPtr<CrossTensor>>;
``` 

Where index has the following signature:

```c++
std::shared_ptr<CrossTensor> index(
    const std::shared_ptr<CrossTensor> &tensor,
    rust::Slice<const std::shared_ptr<CrossTensor>&> indices);
```

When I try to compile the library, it seems that the code generator is producing a bridge signature with duplicate `const` keywords, which causes compilation errors:

```c++
::rust::repr::PtrLen cxxbridge1$index(const ::std::shared_ptr<::CrossTensor> &tensor, ::rust::Slice<const const ::std::shared_ptr<::CrossTensor> &> indices, ::std::shared_ptr<::CrossTensor> *return$) noexcept 
```

I don't know if the `&[&SharedPtr<CrossTensor>]` declaration is malformed or if I should not use a slice in this case. I would appreciate any help with this issue.

dtolnay:
This looks similar to #750. We will need to generate something like `rust::Slice<const std::shared_ptr<CrossTensor> &const>` (I think...).

- One `const` is the one in `const shared_ptr<T> &` corresponding to your `&SharedPtr<T>`;
- The other `const` is the one in `rust::Slice<const T>` corresponding to your `&[T]`.

andfoy:
I've just opened https://github.com/dtolnay/cxx/pull/831, that should fix the duplicate const const

PR #831, opened 2021-04-13
Prevent emitting const const on C++ codegen when using a Slice with a reference

andfoy:
Fixes https://github.com/dtolnay/cxx/issues/830

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -1207,17 +1207,15 @@ fn write_type(out: &mut OutFile, ty: &Type) {
             if slice.mutability.is_none() {
                 match slice.inner {
                     Type::Ref(_) => {
-                        write_type(out, &slice.inner);
-                        write!(out, "const");
+                        write!(out, "");
                     }
                     _ => {
                         write!(out, "const ");
                         write_type(out, &slice.inner);
                     }
                 }
-            } else {
-                write_type(out, &slice.inner);
             }
+            write_type(out, &slice.inner);
             write!(out, ">");
         }
         Type::Fn(f) => {
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -1208,7 +1208,7 @@ fn write_type(out: &mut OutFile, ty: &Type) {
                 match slice.inner {
                     Type::Ref(_) => {
                         write_type(out, &slice.inner);
-                        write!(out, "const")
+                        write!(out, "const");
                     }
                     _ => {
                         write!(out, "const ");
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -1205,7 +1205,14 @@ fn write_type(out: &mut OutFile, ty: &Type) {
         Type::SliceRef(slice) => {
             write!(out, "::rust::Slice<");
             if slice.mutability.is_none() {
-                write!(out, "const ");
+                match slice.inner {
+                    Type::Ref(_) => {
+                        write!(out, "");
+                    }
+                    _ => {
+                        write!(out, "const ");
+                    }
+                }
             }
             write_type(out, &slice.inner);
             write!(out, ">");
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -1207,14 +1207,17 @@ fn write_type(out: &mut OutFile, ty: &Type) {
             if slice.mutability.is_none() {
                 match slice.inner {
                     Type::Ref(_) => {
-                        write!(out, "");
+                        write_type(out, &slice.inner);
+                        write!(out, "const")
                     }
                     _ => {
                         write!(out, "const ");
+                        write_type(out, &slice.inner);
                     }
                 }
+            } else {
+                write_type(out, &slice.inner);
             }
-            write_type(out, &slice.inner);
             write!(out, ">");
         }
         Type::Fn(f) => {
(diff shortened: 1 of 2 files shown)


gen/src/write.rs
```diff
@@ -1205,7 +1205,14 @@ fn write_type(out: &mut OutFile, ty: &Type) {
         Type::SliceRef(slice) => {
             write!(out, "::rust::Slice<");
             if slice.mutability.is_none() {
-                write!(out, "const ");
+                match slice.inner {
+                    Type::Ref(_) => {
+                        write!(out, "");
```
andfoy, line 1210: @dtolnay Should I also add the const write here?

gen/src/write.rs
```diff
@@ -1205,9 +1205,19 @@ fn write_type(out: &mut OutFile, ty: &Type) {
         Type::SliceRef(slice) => {
             write!(out, "::rust::Slice<");
             if slice.mutability.is_none() {
-                write!(out, "const ");
+                match slice.inner {
+                    Type::Ref(_) => {
```
andfoy, line 1209: @dtolnay Do I need to consider other types here?

dtolnay: Yeah slices of pointers are still affected:

```console
$ echo '#[cxx::bridge] mod ffi { extern "Rust" { fn f(x: &[*const u8]); }}' | cargo run --manifest-path gen/cmd/Cargo.toml -- --header - | tail -1

void f(::rust::Slice<const const ::std::uint8_t *> x) noexcept;
```

dtolnay (changes requested):
This turns out not to fix #830. :confused: 

```cpp
#include "rust/cxx.h"
void f(::rust::Slice<const ::std::uint8_t &const> x) noexcept;
```

```console
$ clang++-11 -std=c++20 main.cc -Wall
main.cc:2:44: error: 'const' qualifier may not be applied to a reference
void f(::rust::Slice<const ::std::uint8_t &const> x) noexcept;
                                           ^
$ g++-10 -std=c++20 main.cc -Wall
main.cc:2:44: error: ‘const’ qualifiers cannot be applied to ‘const uint8_t&’ {aka ‘const unsigned char&’}
  2 | void f(::rust::Slice<const ::std::uint8_t &const> x) noexcept;
    |                                            ^~~~~
```

andfoy:
It seems that a container cannot hold references either: https://stackoverflow.com/questions/1155142/why-do-i-get-an-error-in-forming-reference-to-reference-type-map, instead one should wrap a reference around `std::reference_wrapper`, however, I don't know how to extend Slice to convert a reference into a `reference_wrapper`

dtolnay:
Maybe with a [partial template specialization](https://en.cppreference.com/w/cpp/language/partial_specialization) in cxx.h which applies to instantiations of `rust::Slice<T &>`?

```cpp
template <typename T>
class Slice<T &> final {
    /*...*/
};
```

Alternatively it's possible we can fix this purely by using the existing partial specialized helpers that already exist in `<type_traits>`. For example instead of:

https://github.com/dtolnay/cxx/blob/064cf12b7850e4a3e2e09134eeea8fc619e60792/include/cxx.h#L170

we'd have:

```diff
-   T &operator[](std::size_t n) const noexcept;
+   std::add_lvalue_reference<T>::type operator[](std::size_t n) const noexcept;
```

According to https://en.cppreference.com/w/cpp/types/add_reference that has the semantics we want.

> If `T` is a function type that has no cv- or ref- qualifier or an object type, provides a member typedef `type` which is `T&`. If `T` is an rvalue reference to some type `U`, then `type` is `U&`. Otherwise, `type` is `T`.

In other words `rust::Slice<const Struct>::operator[]` would return `const Struct &`, whereas `rust::Slice<const Struct &>::operator[]` would *also* return `const Struct &`, because add_lvalue_reference is instantiated with a type that is already a reference.

andfoy:
> In other words rust::Slice<const Struct>::operator[] would return const Struct &, whereas rust::Slice<const Struct &>::operator[] would also return const Struct &, because add_lvalue_reference is instantiated with a type that is already a reference.

Sounds like it would fix the issue, let me update the PR!

andfoy:
I think it is necessary to update the generated cxx.cc file to consider `add_lvalue_reference` 

```
error: no declaration matches ‘T& rust::cxxbridge1::Slice<T>::operator[](std::size_t) const
```

andfoy:
I tried this change locally, and it seems to work, although I have to use `std::reference_wrapper` on the signature

```rust
match &slice.inner {
    Type::Ref(r) => {
        // let inner_ref = r.inner;
        write!(out, "std::reference_wrapper<");
        write_type(out, &r.inner);
        write!(out, ">");
    }
    _ => {
        write!(out, "const ");
        write_type(out, &slice.inner);
    }
}
```

```c++
std::shared_ptr<CrossTensor> index(
    const std::shared_ptr<CrossTensor> &tensor,
    rust::Slice<std::reference_wrapper<CrossTensorRef>> indices);
```

andfoy:
@dtolnay, do you prefer using `std::reference_wrapper` here?

gen/src/write.rs
sara192: > @dtolnay Do I need to consider other types here?

You

TheZoq2:
Any chance of this being merged? I ran into the issue described in #830 so having a fix would be nice. I'd be happy to help out if some work is needed to get it done :)

PR #834, opened 2021-04-14
Support inner attributes inside bridge

### LINKED ISSUE #833: Bridge attribute breaks inner attributes
```rust
#[cxx::bridge]
mod ffi {
    #![deny(missing_docs)]

    unsafe extern "C++" {

        /// wow
        fn f();
    }
}

pub use self::ffi::*;
```

```console
error: an inner attribute is not permitted in this context
 --> src/main.rs:3:5
  |
3 |     #![deny(missing_docs)]
  |     ^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: inner attributes, like `#![no_std]`, annotate the item enclosing them, and are usually found at the beginning of source files. Outer attributes, like `#[test]`, annotate the item following them.
```

Probably what's going wrong is we are accidentally transporting the inner attributes out to outside of the corresponding item in the macro output, which is not where they are supposed to go.

dtolnay:
Fixes #833.

Example:

```rust
#[cxx::bridge]
mod ffi {
    #![deny(missing_docs)]

    unsafe extern "C++" {

        /// wow
        fn f();
    }
}

pub use self::ffi::*;
```

### DIFF
--- a/syntax/attrs.rs
+++ b/syntax/attrs.rs
@@ -224,7 +224,19 @@ impl OtherAttrs {
 impl ToTokens for OtherAttrs {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         for attr in &self.0 {
-            attr.to_tokens(tokens);
+            let Attribute {
+                pound_token,
+                style,
+                bracket_token,
+                path,
+                tokens: attr_tokens,
+            } = attr;
+            pound_token.to_tokens(tokens);
+            let _ = style; // ignore; render outer and inner attrs both as outer
+            bracket_token.surround(tokens, |tokens| {
+                path.to_tokens(tokens);
+                attr_tokens.to_tokens(tokens);
+            });
         }
     }
 }


dtolnay (commented):
This implementation avoids having to change all the various places that #attrs are inserted into macro output.

https://github.com/dtolnay/cxx/blob/1a9675a7375eee417bc029bbe0d8145f81b11f9b/macro/src/expand.rs#L132-L136
https://github.com/dtolnay/cxx/blob/1a9675a7375eee417bc029bbe0d8145f81b11f9b/macro/src/expand.rs#L167-L171
https://github.com/dtolnay/cxx/blob/1a9675a7375eee417bc029bbe0d8145f81b11f9b/macro/src/expand.rs#L323-L327

All of those would otherwise have to be changed something like:

```rust
let outer_attrs = attrs.filter(|attr| attr.style.is_outer());
let inner_attrs = attrs.filter(|attr| attr.style.is_inner());

quote! {
    #outer_attrs
    mod #ident {
        #inner_attrs
        ...
    }
}
```

which is annoying and unnecessary.

PR #882, opened 2021-06-07
add additional cmake example

trondhe:
Added an additional example on how to integrate cxx with a cmake project, this utilizing Corrosion.


dtolnay (approved):
Thanks!

PR #893, opened 2021-07-01
bazel: don't use cargo in $PATH, make the build hermetic.

ccontavalli:
Background:
tools/bazel/vendor is just running 'cargo', hoping to find it somewhere
in $PATH.

Problem:
rules_rust try to be hermetic, and not use the cargo of the system, but
the cargo of the toolchain installed.
This means that if the cargo on the system is different from the cargo
used by cxx.rs, or the system does not have a cargo at all, bad things
will happen when using cxx.rs.

In this change:
use the recommended way to find/install the tool. Thanks to Andre in the

### DIFF
--- a/tools/bazel/vendor.bzl
+++ b/tools/bazel/vendor.bzl
@@ -2,6 +2,8 @@
 of a crate in the current workspace.
 """
 
+load("@rules_rust//rust:repositories.bzl", "load_arbitrary_tool", "DEFAULT_RUST_VERSION")
+
 def _impl(repository_ctx):
     # Link cxx repository into @third-party.
     lockfile = repository_ctx.path(repository_ctx.attr.lockfile)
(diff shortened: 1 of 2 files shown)


dtolnay (changes requested):
Thanks!

Could you look into the failure in CI? This build command appears to be failing:

https://github.com/dtolnay/cxx/blob/d0d1adf72ae4bfb38b872adbf876453ea515f40b/.github/workflows/ci.yml#L93

https://github.com/dtolnay/cxx/runs/2959406033

```console
ERROR: Failed to load Starlark extension '@rules_rust//rust:repositories.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @rules_rust
This could either mean you have to add the '@rules_rust' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cycles detected during target parsing
INFO: Elapsed time: 0.370s
INFO: 0 processes.
ERROR: Build failed. Not running target
```

tools/bazel/vendor.bzl
```diff
@@ -54,6 +75,13 @@ def _log_cargo_vendor(repository_ctx, result):
 vendor = repository_rule(
     doc = "A rule used to vendor the dependencies of a crate in the current workspace",
     attrs = {
+        "cargo_version": attr.string(
```
UebelAndre, line 78: I think you should also add `cargo_iso_version` to complete the api of `load_arbitrary_tool`.

tools/bazel/vendor.bzl
```diff
@@ -54,6 +75,13 @@ def _log_cargo_vendor(repository_ctx, result):
...
+            doc = "The version of cargo to use",
+            default = "1.51.0",
```
UebelAndre, line 80: Perhaps default this to `DEFAULT_RUST_VERSION` from `@rules_rust//rust:repositories.bzl`

UebelAndre (changes requested):
I also think it'd be a good idea to update the WORKSPACE file to specify the same version of cargo the toolchains are using (as defined by the `rust_repositories` call).

Otherwise seems like a great change! 😄

UebelAndre:
re https://github.com/dtolnay/cxx/pull/893#pullrequestreview-696839269

The vendor rule will need to be loaded after `rules_rust` is defined. In general, it's good to load rules just before you use them in a Bazel WORKSPACE file. (for those not super familiar with Bazel).
```python
load("//tools/bazel:vendor.bzl", "vendor")
```

ccontavalli:
@UebelAndre thanks for the great suggestions! Should all be addressed in the updated commit I just pushed.

@dtolnay tests are passing locally now. As @UebelAndre it was just a matter of ordering the instructions so rules_rust would be loaded before it was used.

UebelAndre (approved):
Looks good to me!

dtolnay (approved):
Thanks!

PR #897, opened 2021-07-06
Issue 87: Bindings for Option

kuecks:
Added bindings to rust's std::option::Option, and C++'s std::optional::optional. Still a WIP but this is able to compile and run a simple example


src/cxx_optional.rs
```diff
@@ -0,0 +1,387 @@
...
+use core::mem::{self, MaybeUninit};
+use core::pin::Pin;
+
+/// Binding to C++ `std::optional<T>`.
+///
+/// # Invariants
+///
+/// As an invariant of this API and the static analysis of the cxx::bridge
+/// macro, in Rust code we can never obtain a `CxxOptional` by value. Instead in
+/// Rust code we will only ever look at a vector behind a reference or smart
```
cgwalters, line 18: s/vector/optional/

cgwalters (commented):
I only superficially skimmed this (and am not a reviewer here really) - mostly just wanted to say thanks for working on this!  My codebase extensively uses `Option<T>` and this will be really nice to have.

CAD97:
I just want to note that this (deliberately) conflicts with #868, in which I would like to have `Option<&T>`<sub>Rust</sub> translate to `T const*`<sub>C++</sub> on the C++ side, rather than `::rust::Option<T const*>`<sub>C++</sub> as implemented here. This is specifically for binding to existing C++ APIs that accept `T*` as either a valid pointer or `nullptr`.

I also ran into the issue of `Option<Pin<&mut T>>` versus `Pin<Option<&mut T>>` and also of `Option<Option<&T>>` in that PR (`Pin<Pin<&mut T>>` is allowed today and normalizes to just `Pin<&mut T>`); so make sure that you don't have the same issue here. I don't _think_ you do, since I modeled `Option` as part of the pointer type (like `Pin`) and you've modeled `Option` as a new kind of type (like `Box`), but it's worth double checking that nesting works properly.

kuecks:
> I also ran into the issue of `Option<Pin<&mut T>>` versus `Pin<Option<&mut T>>` and also of `Option<Option<&T>>` in that PR (`Pin<Pin<&mut T>>` is allowed today and normalizes to just `Pin<&mut T>`); so make sure that you don't have the same issue here.

Thanks for mentioning `Pin`, I wasn't handling it at all and needed to add special case for it.

kuecks:
Notes: all the changes under demo/ will be reverted, that's just my playground. The code feels quire repetitive and nasty, but I can't figure out any better way, there's just so many subtly different cases :(

src/rust_option.rs
```diff
@@ -0,0 +1,230 @@
...
+            None
+        }
+    }
+
+    pub unsafe fn into_inner_unchecked(mut self) -> T {
+        ManuallyDrop::take(&mut self.inner.value)
+    }
+}
+
+impl<'a, T> RustOption<&'a T> {
```
kuecks, line 40: I think some of these implementations can be removed but maybe its future proofing to have them? I can try to see which ones are unneeded

src/rust_option.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub unsafe fn into_option_mut_improper<'a, T>(self) -> Option<&'a mut T> {
+        std::mem::transmute::<RustOption<*mut std::ffi::c_void>, Option<&'a mut T>>(self)
+    }
+
+    pub unsafe fn as_mut_option_mut_improper<'a, T>(&mut self) -> &mut Option<&'a mut T> {
+        &mut *(self as *mut RustOption<*mut std::ffi::c_void> as *mut Option<&'a mut T>)
+    }
+}
+
+impl RustOption<crate::private::RustStr> {
```
kuecks, line 208: These seemed particularly sketchy

syntax/resolve.rs
```diff
@@ -13,7 +13,9 @@ impl<'a> Types<'a> {
         let ident = ident.ident();
         match self.try_resolve(ident) {
             Some(resolution) => resolution,
-            None => panic!("Unable to resolve type `{}`", ident),
+            None => {
```
kuecks, line 16: hrmm i think this should be reverted

djpiper28:
Is this pr going anywhere? I am quite interested in this feature.

kuecks:
@djpiper28 yes I am still interested in pushing this through, I think dtolnay is (understandably) just quite busy unfortunately. I may explore alternative implementations, specifically reference_wrapper, in the meantime

VirxEC:
Any progress? This feature would be pretty big...

xdray:
is anyone doing this?

JanKaul:
Maybe it would be easier to merge this PR if it is decomposed into smaller parts. If I understand correctly this PR adds support for rust `Option<T>` as well as introduces `CxxOptional` for C++ `std::optional`. Maybe we can split those up into two separate PRs. We can then focus on the one that is simpler to implement.

It looks like the memory layout of rust's `Option<T>` adds some complexity on the C++ side. On the other hand `std::optional` seems to have a fixed memory layout and might be easier to implement.

Issue #835, opened 2021-04-14
Repr alignment attribute on shared structs


dtolnay:
```rust
#[cxx::bridge]
mod ffi {
    #[repr(align(4))]
    struct S {
        b: [u8; 4],
    }
}
```

Expected C++ header:

```cpp
struct alignas(4) S {
  std::array<std::uint8_t, 4> b;
};
```

- Rust reference: https://doc.rust-lang.org/1.51.0/reference/type-layout.html#the-alignment-modifiers
- C++ reference: https://en.cppreference.com/w/cpp/language/alignas

<br>

Currently the C++ code generator just errors on this attribute.

```console
error[cxxbridge]: unrecognized repr
  ┌─ src/main.rs:3:12
  │
3 │     #[repr(align(4))]
  │            ^^^^^^^^ unrecognized repr
```

PR #902, opened 2021-07-21
Add support for `#[repr(align(x))]` on bridge structs

mxxo:
Add support for `#[repr(align(x))]` on bridge structs, closing #835. For the example code: 
```rust 
#[cxx::bridge]
mod ffi {
    #[repr(align(4))]
    struct S {
        b: [u8; 4],
    }
}
```
the following output was obtained using the cxx code generators:
```rust 
// -- snip 
#[repr(C)]
#[repr(align(4))]
pub struct S {
    pub b: [u8; 4],
}
// -- snip
```
```cpp
#include <array>
#include <cstdint>
#include <type_traits>

struct S;

#ifndef CXXBRIDGE1_STRUCT_S
#define CXXBRIDGE1_STRUCT_S
struct alignas(4) S final {
  ::std::array<::std::uint8_t, 4> b;

  using IsRelocatable = ::std::true_type;
};
#endif // CXXBRIDGE1_STRUCT_S
```
I believe it would be straightforward to extend this implementation for `#[repr(packed)]` as well. 

I was wondering whether to add support for enums but decided against it for now and made it an "unsupported" error for the time being. `enum alignas(X)` is accepted by all the C++ compilers I tried, but I also found a defect report from 2017 (2534) that removes the ability to apply `alignas` to enums: https://wg21.cmeerw.net/cwg/issue2354.

### DIFF
--- a/syntax/parse.rs
+++ b/syntax/parse.rs
@@ -2,9 +2,10 @@ use crate::syntax::attrs::OtherAttrs;
 use crate::syntax::discriminant::DiscriminantSet;
 use crate::syntax::file::{Item, ItemForeignMod};
 use crate::syntax::report::Errors;
+use crate::syntax::repr::Repr;
 use crate::syntax::Atom::*;
 use crate::syntax::{
-    attrs, error, Api, Array, Derive, Doc, Enum, EnumRepr, ExternFn, ExternType, ForeignName, Impl,
+    attrs, error, Alignment, Api, Array, Derive, Doc, Enum, EnumRepr, ExternFn, ExternType, ForeignName, Impl,
     Include, IncludeKind, Lang, Lifetimes, NamedType, Namespace, Pair, Ptr, Receiver, Ref,
     Signature, SliceRef, Struct, Ty1, Type, TypeAlias, Var, Variant,
 };
@@ -57,6 +58,7 @@ pub fn parse_items(
 fn parse_struct(cx: &mut Errors, mut item: ItemStruct, namespace: &Namespace) -> Result<Api> {
     let mut doc = Doc::new();
     let mut derives = Vec::new();
+    let mut repr = None;
     let mut namespace = namespace.clone();
     let mut cxx_name = None;
     let mut rust_name = None;
@@ -66,12 +68,19 @@ fn parse_struct(cx: &mut Errors, mut item: ItemStruct, namespace: &Namespace) ->
         attrs::Parser {
             doc: Some(&mut doc),
             derives: Some(&mut derives),
+            repr: Some(&mut repr),
             namespace: Some(&mut namespace),
             cxx_name: Some(&mut cxx_name),
             rust_name: Some(&mut rust_name),
             ..Default::default()
         },
     );
+
+    let alignment = if let Some(Repr::Align(x)) = repr {
+        Some(Alignment::Align(x))
+    } else {
+        None
+    };
 
     let named_fields = match item.fields {
         Fields::Named(fields) => fields,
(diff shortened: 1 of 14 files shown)


dyangelo-grullon:
@dtolnay is there any feedback on this change? What is the path to getting this merged? I’m contemplating closing the last mile or putting up a bounty for this specific feature.

syntax/parse.rs
```diff
@@ -214,6 +224,15 @@ fn parse_enum(cx: &mut Errors, item: ItemEnum, namespace: &Namespace) -> Api {
         cx.error(where_clause, "enum with where-clause is not supported");
     }
 
+    let repr = match repr {
+        Some(Repr::Atom(atom)) => Some(atom),
+        Some(Repr::Align(_)) => {
+            cx.error(&item, "repr(align) on enums is not supported");
```
dtolnay, line 230: It should be clarified that this is a limitation of C++ enums and not cxx.

```suggestion
            cx.error(&item, "C++ does not support custom alignment on an enum");
```

syntax/repr.rs
```diff
@@ -0,0 +1,45 @@
+use crate::syntax::Atom::{self, *};
+use syn::parse::{Error, Parse, ParseStream, Result};
+use syn::{Ident, LitInt};
+
+#[derive(Copy, Clone, PartialEq)]
```
dtolnay, line 5: None of these trait impls are used anywhere.

```suggestion
```

syntax/repr.rs
```diff
@@ -0,0 +1,45 @@
...
+            match atom {
+                U8 | U16 | U32 | U64 | Usize | I8 | I16 | I32 | I64 | Isize if input.is_empty() => {
+                    return Ok(Repr::Atom(atom));
+                }
+                _ => {}
+            }
+        } else if ident == "align" {
+            let content;
+            syn::parenthesized!(content in input);
+            let alignment: u32 = content.parse::<LitInt>()?.base10_parse()?;
```
dtolnay, line 25: This would accept a suffixed integer literal such as `#[repr(align(2int))]`. That should not be allowed.

syntax/repr.rs
```diff
@@ -0,0 +1,45 @@
...
+            if !alignment.is_power_of_two() {
+                return Err(Error::new_spanned(
+                    begin.token_stream(),
+                    "invalid `repr(align)` attribute: not a power of two",
+                ));
+            }
+            if alignment > 2u32.pow(29) {
```
dtolnay, line 32: I don't think there is a guarantee that C++ compilers support alignments in this whole range. For example Clang:

```cpp
struct alignas(1<<19) A {};
struct alignas(1<<29) B {};
auto a() { return alignof(A); } // 524288
auto b() { return alignof(B); } // 1
```

This needs to use a smaller limit until somebody needs otherwise. As is, this could lead to C++ passing an insufficiently aligned reference to Rust which would be UB.

syntax/repr.rs
```diff
@@ -0,0 +1,45 @@
...
+        let begin = input.cursor();
+        let ident: Ident = input.parse()?;
+        if let Some(atom) = Atom::from(&ident) {
+            match atom {
+                U8 | U16 | U32 | U64 | Usize | I8 | I16 | I32 | I64 | Isize if input.is_empty() => {
+                    return Ok(Repr::Atom(atom));
+                }
+                _ => {}
+            }
+        } else if ident == "align" {
```
dtolnay, line 22: It would be helpful to give a better error message if a repr attribute is written with a non-literal expression, since C++ allows this but Rust does not. `#[repr(align(1 << 6))]`

syntax/parse.rs
dtolnay: Also the error should point to just the align attribute, not the entire item.

PR #906, opened 2021-07-31
Add UTF-16 rust::String initialization

Gerharddc:
This patch exposes String::from_utf16 to the C++ interface for rust::String.

### DIFF
--- a/src/cxx.cc
+++ b/src/cxx.cc
@@ -31,8 +31,10 @@ void cxxbridge1$cxx_string$push(std::string &s, const std::uint8_t *ptr,
 void cxxbridge1$string$new(rust::String *self) noexcept;
 void cxxbridge1$string$clone(rust::String *self,
                              const rust::String &other) noexcept;
-bool cxxbridge1$string$from(rust::String *self, const char *ptr,
-                            std::size_t len) noexcept;
+bool cxxbridge1$string$from_utf8(rust::String *self, const char *ptr,
+                                 std::size_t len) noexcept;
+bool cxxbridge1$string$from_utf16(rust::String *self, const char16_t *ptr,
+                                  std::size_t len) noexcept;
 void cxxbridge1$string$drop(rust::String *self) noexcept;
 const char *cxxbridge1$string$ptr(const rust::String *self) noexcept;
 std::size_t cxxbridge1$string$len(const rust::String *self) noexcept;
@@ -81,8 +83,14 @@ String::String(String &&other) noexcept : repr(other.repr) {
 String::~String() noexcept { cxxbridge1$string$drop(this); }
 
 static void initString(String *self, const char *s, std::size_t len) {
-  if (!cxxbridge1$string$from(self, s, len)) {
+  if (!cxxbridge1$string$from_utf8(self, s, len)) {
     panic<std::invalid_argument>("data for rust::String is not utf-8");
+  }
+}
+
+static void initString(String *self, const char16_t *s, std::size_t len) {
+  if (!cxxbridge1$string$from_utf16(self, s, len)) {
+    panic<std::invalid_argument>("data for rust::String is not utf-16");
   }
 }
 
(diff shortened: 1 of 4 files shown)


tests/ffi/tests.cc
```diff
@@ -841,6 +841,12 @@ extern "C" const char *cxx_run_test() noexcept {
   ASSERT(cstr == "foo");
   ASSERT(other_cstr == "test");
 
+  const char* utf8_literal = u8"Test string";
+  const char16_t* utf16_literal  = u"Test string";
+  rust::String utf8_rstring = utf8_literal;
+  rust::String utf16_rstring = utf16_literal;
+  ASSERT(utf8_rstring == utf16_rstring);
```
dtolnay, line 848: Do you have any insight into why the test suite failed with MSVC on 29b623436fca88aa8af3b3fd5445dd5733ef48b2?

```console
Assertion failed: `utf8_rstring == utf16_rstring`, tests.cc:848
thread 'test_c_call_r' panicked at 'cxx_run_test()', tests\test.rs:238:5
```

Guessing it was the missing `u8` on the string literal right? Unannotated string literal has implementation-defined system encoding whereas `u8"..."` is specified to UTF-8.

But I see that you also removed a multi-byte code point from the test string in between. Why was that?

tests/ffi/tests.cc
```diff
@@ -841,6 +841,12 @@ extern "C" const char *cxx_run_test() noexcept {
   ASSERT(cstr == "foo");
   ASSERT(other_cstr == "test");
 
+  const char* utf8_literal = u8"Test string";
+  const char16_t* utf16_literal  = u"Test string";
```
dtolnay, line 845:
```suggestion
  const char16_t* utf16_literal = u"Test string";
```

src/cxx.cc
```diff
@@ -33,6 +33,8 @@ void cxxbridge1$string$clone(rust::String *self,
                              const rust::String &other) noexcept;
 bool cxxbridge1$string$from(rust::String *self, const char *ptr,
```
dtolnay, line 34: Please rename this symbol now that it's no longer the unique way to construct a string from data.

```suggestion
bool cxxbridge1$string$from_utf8(rust::String *self, const char *ptr,
```

src/cxx.cc
```diff
@@ -100,6 +108,18 @@ String::String(const char *s, std::size_t len) {
...
 
+String::String(const char16_t *s) {
+  assert(s != nullptr);
+  initString(this, s, std::char_traits<char16_t>::length(s));
+}
+
+String::String(const char16_t *s, std::size_t len) {
+  assert(s != nullptr || len == 0);
+  initString(this,
+             s == nullptr && len == 0 ? reinterpret_cast<const char16_t *>(1) : s,
```
dtolnay, line 119: This is UB because 1 is not a legal pointer value for the \&\[u16\] unsafely constructed by slice::from_raw_parts on the Rust side. See the documentation in https://doc.rust-lang.org/1.54.0/std/slice/fn.from_raw_parts.html#safety about alignment.

dtolnay (changes requested):
Thanks for the PR! Mostly looks good to me.

tests/ffi/tests.cc
Gerharddc: Yeah the original failure was caused by the missing `u8` but when I started debugging a bit more I saw that multi-byte code points still don't print correctly to the console on MSVC even though the UTF-8 and UTF-16 strings then match. To confirm that the issue is not related to the patch I also tried printing the raw literal to the console and got the same garbage. To me this sounds like some peculiarity with string literals on MSVC rather than an issue with the patch itself so I decided to remove the code point rather than to create the illusion that defining them in literals like that works on MSVC.

P.S. Looking around it seems the issue might not be with MSVC, but rather with the Windows console not accepting UTF-8. Either way I couldn't get it printing correctly and I don't care enough about MSVC at the moment to figure out exactly what the issue is.

tests/ffi/tests.cc
Gerharddc: Sorry about that one, forgot to run clang-format

src/cxx.cc
Gerharddc: Done

src/cxx.cc
Gerharddc: Interesting, thanks for pointing that out. I changed to it 2, guessing that is correct alignment for a slice with 2 byte long elements?

dtolnay (approved):
Nice -- thank you!

PR #909, opened 2021-08-04
Add let_cxx_vector to construct CxxVector on stack

compiler-errors:
Implement `let_cxx_vector` macro, very similarly to `let_cxx_string`, to enable constructing a `CxxVector<T>` on the stack. 

Helps with calling methods that take `std::vector` as an out variable.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -1854,6 +1854,43 @@ fn write_cxx_vector(out: &mut OutFile, key: NamedImplKey) {
     writeln!(out, "  return &(*s)[pos];");
     writeln!(out, "}}");
 
+    writeln!(
+        out,
+        "  static_assert(alignof(std::vector<{}>) <= alignof(void *),",
+        inner
+    );
+    writeln!(
+        out,
+        "    \"unexpectedly large std::vector<{}> alignment\");",
+        inner
+    );
+    writeln!(
+        out,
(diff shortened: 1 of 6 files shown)


compiler-errors:
Oops, had a dumb naming issue. Strange that the tests were only failing on windows targets, though. :sweat:

compiler-errors:
@hanusek's #961 reminded me that this PR was open.

@dtolnay: is this change something you consider reasonable, and if so, can I get a review on it sometime? If it's not worthwhile or is fundamentally broken in some way, I'll just close it.

NlightNFotis:
Hi @dtolnay, pinging you here, as evaluating this one for going in would be an incredible help to a lot of workflows (at least for me).

PR looks clean, has tests, could do with a rebase on top of recent `master` but no merge conflicts.

What would it take for this to be taken forward?

NlightNFotis:
Hi @dtolnay, many thanks for a very prompt response.

In light of this decision, what would you suggest is the best way to construct a `CxxVector<CxxString>` and pass it to the C++ side?

Would I be better off just creating the Rust equivalents and translate them into the C++ APIs in the `.cpp` file?

Thanks again,

dtolnay:
You can do that or you can obtain an empty UniquePtr\<CxxVector\<CxxString\>\> from C++ to put elements into. Or, via a callback, obtain a Pin\<\&mut CxxVector\<CxxString\>\> for the Rust code to fill.

PR #947, opened 2021-10-05
Pull in newer Bazel rules_rust

dtolnay:
`DEFAULT_RUST_VERSION` no longer exists on upstream's main branch.


dtolnay:
That should work &mdash; we'll just link using `lld` which is not order sensitive to link args.

PR #951, opened 2021-10-20
Add Vec clear

### LINKED ISSUE #950: How do you clear a Vec<T> from C++?
I have a shared struct that contains a Rust Vec.

``` Rust
struct Foo {
    values: Vec<f32>
}
```

This becomes a `rust::cxxbridge::Vec` on the C++ side.
Unfortunately this class doesn't seem to have a `clear` method which means I cannot re-use the same vector.

How can I clear an existing `Vec` from C++?

rookboom:
Added a `clear` method to rust::Vec to address https://github.com/dtolnay/cxx/issues/950.

Closes #950.

### DIFF
--- a/book/src/binding/vec.md
+++ b/book/src/binding/vec.md
@@ -45,6 +45,7 @@ public:
   void reserve(size_t new_cap);
   void push_back(const T &value);
   void push_back(T &&value);
+  void clear();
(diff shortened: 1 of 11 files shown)


dtolnay (approved):
Thanks!

PR #963, opened 2021-11-27
Docs: required info to build via Bazel in external repo

jon-chuang:


dtolnay (commented):
Thanks for the PR!

I don't immediately understand why this would be necessary. For example, rules_rust's [examples/crate_universe/using_cxx](https://github.com/bazelbuild/rules_rust/tree/959db8ed3cea851fc12cf24bc76c2ea733fd9e05/examples/crate_universe/using_cxx) uses cxx without lld, and seems to work.

bsilver8192:
I suspect it depends on your toolchain and which `cxx` features you use. Some toolchains will use a linker that doesn't care about the order by default, and some very minimal uses of `cxx` don't have the circular linking order either. #1069 adds more explicit documentation about the problem.

PR #1005, opened 2022-01-25
Add UniquePtr::to_shared and SharedPtr::from_unmanaged

russelltg:
This adds a way to create SharedPtr's from C++ pointers created with `new` (or via `unique_ptr`s).

I'm a new contributor here, let me know if I did this in a reasonable way, and I'm happy to put some effort in if this is something that's interesting.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -1801,6 +1801,18 @@ fn write_shared_ptr(out: &mut OutFile, key: NamedImplKey) {
     begin_function_definition(out);
     writeln!(
         out,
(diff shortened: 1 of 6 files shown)


russelltg:
I fixed that error and added an additional test to cover the situation, but it needs approval to run again. Just pinging just in case you don't get commit notifications.

russelltg:
Freshly rebased onto master :)

russelltg:
Hmm I think it would be reasonable also to implement `SharedPtr<T> from UniquePtr<T>`, happy to add that if you agree.

PR #1012, opened 2022-02-12
Change base name for cxx_build object file according to source file name

rodrigodesalvobraz:
Per https://github.com/dtolnay/cxx/issues/1010#issuecomment-1034190083


dtolnay (commented):
I don't think there is a need for `cxx`-specific guidance on that call. To the extent that guidance is needed, I would prefer for it to be in the documentation of `cc::Build::compile` in the cc crate. We can make the cxx tutorial link to that documentation instead.

I'll comment in #1010 explaining why I don't think the cc source file name is a good choice here.

Thanks!

PR #1029, opened 2022-04-02
[bazel] Only use lld on Linux

keith:
macOS C++ toolchains from Apple do not contain lld, so bazel builds in
this repo would fail. This change scopes this only to Linux


dtolnay (approved):
Thanks!

keith:
Thanks!

PR #1037, opened 2022-04-12
[docs] add comment for rust::Str size and length

kassens:
The presence of `size` and `length` confused me, adding a small comment to make it more clear for future readers.


cgwalters (commented):
LGTM

dtolnay (approved):
Thanks!

PR #1038, opened 2022-04-19
support resolving arm triplet with bazel

snowp:
Adds support to vendor.bzl for resolving the arm triplet. This only works with Bazel 5.1, but the way we access the new field should be safe for older versions and result in a negative check, maintaining the old behavior


dtolnay (approved):
Thanks!

Issue #669, opened 2021-01-12
Use /// when propagating doc comments


alexeyr:
In #616 I've noticed `//` is used for the comments; using `///` instead will allow Doxygen to recognize them as doc comments.

gabrik:
It seems not the case. We have some docstrings in the rust part that are indeed as `///` but they are "copied" in the generate files as `//` instead of the format understood by Doxygen [doc](https://www.doxygen.nl/manual/docblocks.html)

It can be useful to have them "copied" with the right format so that Doxygen can be used to generate docs for the bindings.

tomtau:
@dtolnay regarding https://github.com/dtolnay/cxx/pull/1048#discussion_r1020063842
> I understand that it could be simple for someone to work around, but I am not interested in accepting this PR as currently implemented. Taking Rust docs and turning them into either doxygen docs or silently into something that looks like doxygen docs, but isn't, is going to come across as clearly broken behavior to anybody that uses this.

Doxygen has quite a few rules to comply with (e.g. "To document a member of a C++ class, you must also document the class itself."), and having the full validation to ensure a compliant conversion in CXX may be an overkill and passing this responsibility to its users instead is IMHO ok.
But I understand that "feature" shouldn't be called "doxygen" in order not to set wrong expectations. Perhaps what could work is to have an optional `cxx_doc` attribute which could be used to control how comments look like in the generated files, e.g.:
```rust
#[cxx::bridge(namespace = "org::blobstore")]

mod ffi {
    /// Shared structs with fields visible to both languages.
    #[cxx_doc(comment_line = "//!")
    struct BlobMetadata {
...
```
or one could perhaps also use that to have a different comment content:
```rust
#[cxx::bridge(namespace = "org::blobstore")]

mod ffi {
    /// Shared structs with fields visible to both languages.
    #[cxx_doc(comment_line = "///", text = " Shared structs with fields\n visible to both languages.")
    struct BlobMetadata {
...
```

@alexeyr @gabrik what do you think?

dtolnay:
I don't find that an appealing approach.

Regarding the _"to document a member of a C++ class, you must also document the class itself"_ rule, I agree we don't need cxx to check that.

tomtau:
@dtolnay what is your preference? Keep it outside CXX and leave it to custom post-processing scripts to be applied on generated files?

PR #1048, opened 2022-05-04
Add doxygen feature for cxx-build 

damoncro:
Close:
- #669

Solution:
Add `doxygen` feature for `cxx-build`:
Enabling this feature will write `///` comments instead of `//` into generated cpp files.

### DIFF
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -422,7 +422,11 @@ fn write_enum<'a>(out: &mut OutFile<'a>, enm: &'a Enum) {
     writeln!(out, " {{");
     for variant in &enm.variants {
         for line in variant.doc.to_string().lines() {
-            writeln!(out, "  ///{}", line);
+            if cfg!(feature = "doxygen") {
+                writeln!(out, "  ///{}", line);
+            } else {
+                writeln!(out, "  //{}", line);
+            }
         }
         writeln!(out, "  {} = {},", variant.name.cxx, variant.discriminant);
     }
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -236,13 +236,21 @@ fn write_struct<'a>(out: &mut OutFile<'a>, strct: &'a Struct, methods: &[&Extern
     writeln!(out, "#ifndef {}", guard);
     writeln!(out, "#define {}", guard);
     for line in strct.doc.to_string().lines() {
-        writeln!(out, "//{}", line);
+        if cfg!(feature = "doxygen") {
+            writeln!(out, "  ///{}", line);
+        } else {
+            writeln!(out, "  //{}", line);
+        }
     }
     writeln!(out, "struct {} final {{", strct.name.cxx);
 
     for field in &strct.fields {
         for line in field.doc.to_string().lines() {
-            writeln!(out, "  //{}", line);
+            if cfg!(feature = "doxygen") {
+                writeln!(out, "  ///{}", line);
+            } else {
+                writeln!(out, "  //{}", line);
+            }
         }
         write!(out, "  ");
         write_type_space(out, &field.ty);
@@ -256,7 +264,11 @@ fn write_struct<'a>(out: &mut OutFile<'a>, strct: &'a Struct, methods: &[&Extern
             out.next_section();
         }
         for line in method.doc.to_string().lines() {
-            writeln!(out, "  //{}", line);
+            if cfg!(feature = "doxygen") {
+                writeln!(out, "  ///{}", line);
+            } else {
+                writeln!(out, "  //{}", line);
+            }
         }
         write!(out, "  ");
         let sig = &method.sig;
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ std = ["alloc"]
 
 [dependencies]
 cxxbridge-macro = { version = "=1.0.72", path = "macro" }
-link-cplusplus = "1.0"
+link-cplusplus = { version = "1.0", features = ["libc++"] }
```
dtolnay, line 28: Please remove this unrelated change from the PR.

gen/src/write.rs
```diff
@@ -236,13 +236,21 @@ fn write_struct<'a>(out: &mut OutFile<'a>, strct: &'a Struct, methods: &[&Extern
     writeln!(out, "#ifndef {}", guard);
     writeln!(out, "#define {}", guard);
     for line in strct.doc.to_string().lines() {
-        writeln!(out, "//{}", line);
+        if cfg!(feature = "doxygen") {
```
dtolnay, line 239: Could you justify why not to always emit `///`? Is it because that would break anyone currently running Doxygen on cxx-generated headers in which the Rust comments are not "valid Doxygen" in some way?

gen/src/write.rs
```diff
@@ -391,14 +411,18 @@ fn write_enum<'a>(out: &mut OutFile<'a>, enm: &'a Enum) {
...
+            writeln!(out, "  //{}", line);
+        }
     }
     write!(out, "enum class {} : ", enm.name.cxx);
     write_atom(out, repr);
     writeln!(out, " {{");
     for variant in &enm.variants {
         for line in variant.doc.to_string().lines() {
-            writeln!(out, "  //{}", line);
+            writeln!(out, "  ///{}", line);
```
dtolnay, line 425: Missing `if`

gen/src/write.rs
```diff
@@ -337,7 +349,11 @@ fn write_opaque_type<'a>(out: &mut OutFile<'a>, ety: &'a ExternType, methods: &[
     writeln!(out, "#ifndef {}", guard);
     writeln!(out, "#define {}", guard);
     for line in ety.doc.to_string().lines() {
-        writeln!(out, "//{}", line);
+        if cfg!(feature = "doxygen") {
+            writeln!(out, "  ///{}", line);
```
dtolnay, line 353: Wrong indentation.

Elsewhere as well.

gen/src/write.rs
```diff
@@ -236,13 +236,21 @@ fn write_struct<'a>(out: &mut OutFile<'a>, strct: &'a Struct, methods: &[&Extern
     writeln!(out, "#ifndef {}", guard);
     writeln!(out, "#define {}", guard);
     for line in strct.doc.to_string().lines() {
-        writeln!(out, "//{}", line);
+        if cfg!(feature = "doxygen") {
+            writeln!(out, "  ///{}", line);
```
dtolnay, line 240: According to https://www.doxygen.nl/manual/docblocks.html, `///` is only considered Doxygen if there are at least 2 of them on consecutive lines. That's not how Rust `///` works normally so it would make sense for cxx-build to ensure it's converting Rust docs to Doxygen docs.

dtolnay (changes requested):

gen/src/write.rs
damoncro: This PR is not a complete implementation for generating the Doxygen docs. But a small enhancement that is enough for fullfilling the minimal doxygen docs generate requirement.

Two more lines are simple, just input two more lines in rust side.

gen/src/write.rs
damoncro: What do you mean `wrong indentation`, the indentation is already checked by `rustfmt`.

gen/src/write.rs
damoncro: Added

gen/src/write.rs
damoncro: This PR is not a complete implementation for generating the Doxygen docs. But a small enhancement that is enough for fullfilling the minimal doxygen docs generate requirement. If we need more advance docs, maybe it is worth to add more features later on.

gen/src/write.rs
dtolnay: I understand that it could be simple for someone to work around, but I am not interested in accepting this PR as currently implemented. Taking Rust docs and turning them into either doxygen docs or silently into something that _looks_ like doxygen docs, but isn't, is going to come across as clearly broken behavior to anybody that uses this.

gen/src/write.rs
dtolnay: In the generated code, I want the number of spaces in front of the comment to be equal to the number of spaces in front of the thing that it is documenting, like it was prior to this PR.

gen/src/write.rs
damoncro: Noted, let's find a better way to improve it later~

PR #1057, opened 2022-06-13
gen: Fix panic in clap when handling `--include`

cgwalters:
For some reason I don't yet fully understand, the build
of rpm-ostree started failing with a panic in `cxxbridge`:
https://github.com/coreos/rpm-ostree/pull/3763#issuecomment-1154193336

What I haven't worked out is what changed to break this; my best
guess is something like compiler type inference?  AFAICS, the
code here hasn't changed in a long time, and neither has the lockfiles.

(Hmm...unless we're not using lockfiles for this...ah yes, that would
 explain it.  And https://crates.io/crates/clap/3.2.1 was just published.
 Will dig into this more)

Anyways, this is not an elegant fix; we just (correctly) ask clap
to give us `OsString` and not `String`.  The real fix is clearly
to stop using all the deprecated clap APIs, but I'm trying to make
the least invasive change for the moment.

### DIFF
--- a/gen/cmd/src/app.rs
+++ b/gen/cmd/src/app.rs
@@ -60,6 +60,7 @@ const INCLUDE: &str = "include";
 const OUTPUT: &str = "output";
 const CFG: &str = "cfg";
 
+#[allow(deprecated)]
 pub(super) fn from_args() -> Opt {
     let matches = app().get_matches();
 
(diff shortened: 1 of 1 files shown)


cgwalters:
(I briefly looked at trying to parse `--include` as a `PathBuf`, but that can't really work because we write it into generated C code as `#include "<path>"` which...I don't think we want to support non-UTF8 there)

cgwalters:
```
 error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/clap_lex-0.2.2/Cargo.toml`
Caused by:
  feature `edition2021` is required
```

OK well...I guess at this point we probably need to add a `Cargo.lock` and use @dependabot here?

cgwalters:
Trying a different approach over in https://github.com/dtolnay/cxx/pull/1058

cgwalters:
Filed https://github.com/clap-rs/clap/issues/3826

Issue #870, opened 2021-05-09
Throw exception with cause chain


smessmer:
Currently, when a rust ffi function returns a `Result<T>` and that happens to be an error, cxx.rs formats it with `Display` and creates an exception with that message. Unfortunately, that often contains very little information. If using the `anyhow` crate, then this only contains the outermost `context()` message, and doesn't include the chain of causes.

Is there a way to include the chain of causes in the message? I think `anyhow` supports formatting with  `{:#}` to display them.

PR #1063, opened 2022-07-06
Expose the inner rust side error type via the exception on the c++ side

weiznich:
This is an attempt to address https://github.com/dtolnay/cxx/issues/870 and similar issues. 

I attempt to expose more information on the underlying rust error by including the the original error into the c++ side exception. This allows accessing any error related information from the c++ side as well, assuming that the error type is exposed as part of the cxx bridge definition.

I deliberately open this PR in this state, to gather some feedback on the design idea before spending to much time on details.

### DIFF
--- a/gen/src/builtin.rs
+++ b/gen/src/builtin.rs
@@ -42,13 +42,13 @@ impl<'a> Builtins<'a> {
     }
 }
 
-pub(super) fn write(out: &mut OutFile) {
-    if out.builtin == Default::default() {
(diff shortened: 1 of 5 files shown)


weiznich:
Gentle ping to @dtolnay to get some input on the general design and whether this feature is wanted or not.

dtolnay:
I am interested for this crate to support this, but I will not be able to prioritize iterating on the implementation in the near term.

PR #1069, opened 2022-07-17
Add a note about linking order

bsilver8192:
I got pretty far assuming that none of the generated C++ code depends on
the symbols emitted by the procedural macro, before doing something with
Box that requires it. Add a note to help others recognize this pitfall
beforehand.


dtolnay (approved):
Thank you!

PR #1081, opened 2022-08-19
Added documentation for methods on shared types

paandahl:
I only realized methods on shared types were possible by looking through the source code, so I thought it would be a good idea to add some documentation.

I added two sections with code snippets to the "Shared types" page:

- Methods defined in Rust
- Methods defined in C++


ttzytt:
I'm new to the project and also looking for this.  It would be really helpful if it is in the doc.

ttzytt:
Wondering if it is possible to have methods implementing traits on shared types.

PR #1086, opened 2022-08-28
Added another CMake example application

paandahl:
I always wished for an inverted example when learning to use CXX (blobstore in Rust, main() in C++).

I decided to make it, and to keep it as close to the official demo as possible. I also kept the CMake-configuration to a minimum, avoiding advanced topics & tricks, since it's supposed to be an introductory example.

Tested on Linux, Mac, Windows.


dtolnay (approved):
Nice! Thank you.

PR #1090, opened 2022-09-01
[docs] Add `geekbrother/cxx-corrosion-cmake` repo to the CMake docs.

geekbrother:
Update the CMake docs repositories list with the `geekbrother/cxx-corrosion-cmake` repository that has good examples, is easy to build with Docker and Nix, and a modern CMake file to use a CXX+Corrosion.


dtolnay (approved):
Nice :slightly_smiling_face: Thank you!

PR #1091, opened 2022-09-02
Add Unpin impl for UniquePtr.

LeonMatthesKDAB:
Similar to a std::boxed::Box, a UniquePtr itself can be Unpin, even if
its target is not Unpin.
As UniquePtr is not a self-referential type, this is safe to do.

See:
- https://doc.rust-lang.org/std/pin/index.html#unpin
- https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-Unpin

### DIFF
--- a/src/unique_ptr.rs
+++ b/src/unique_ptr.rs
@@ -177,6 +177,10 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #1098, opened 2022-09-19
tests/ffi: strange error with a function returning Vec<Buffer>

naufraghi:
The error triggers a warning for the Buffer field in Array that was already there but somehow silent.

    cargo:warning=.../out/cxxbridge/sources/tests/ffi/lib.rs.cc:1387:45:
        error: static assertion failed: type tests::Buffer should be trivially
               move constructible and trivially destructible in C++ to be used
               as a field of `Array` or vector element in Vec<Buffer> in Rust
    cargo:warning= 1387 |     ::rust::IsRelocatable<::tests::Buffer>::value,

- Here the [failing pipeline log](https://github.com/dtolnay/cxx/actions/runs/3081009896/jobs/4979054704#step:7:57) for the error above (first commit c858291a9ff62bcbd123363407cd73e05ab1a46a).
- The "fix" is not complete either, the plain Buffer struct code does not pass the tests on windows: [msvc pipeline logs](https://github.com/dtolnay/cxx/actions/runs/3081012199/jobs/4979060118#step:7:59) (second commit d0537c91e56e570b880046be092c4068739cb41c)

### DIFF
--- a/tests/ffi/lib.rs
+++ b/tests/ffi/lib.rs
@@ -421,7 +421,9 @@ impl ffi::Array {
 
 #[derive(Default)]
 #[repr(C)]
-pub struct Buffer([c_char; 12]);
+pub struct Buffer {
+    buf: [c_char; 12],
+}
 
 unsafe impl ExternType for Buffer {
     type Id = type_id!("tests::Buffer");
--- a/tests/ffi/lib.rs
+++ b/tests/ffi/lib.rs
@@ -270,6 +270,7 @@ pub mod ffi {
         fn r_return_rust_string() -> String;
         fn r_return_unique_ptr_string() -> UniquePtr<CxxString>;
         fn r_return_rust_vec() -> Vec<u8>;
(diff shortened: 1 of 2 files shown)


naufraghi:
Long story short, I'd like to use a trivial struct inside a `Vec`, but it did not work. I had to wrap it inside a struct as a field to have it work.

```rust
struct Foo {
   data: Vec<TrivialData>
}
```

does not link, missing `Vec` constructor.

```rust
struct DataWrapper {    // <--- shared
   d: TrivialData
}

struct Foo {
   data: Vec<DataWrapper>
}
```

So I dived into the code to understand if that case was covered, tried to add a test and found a strange error.

naufraghi:
I'd like to help, either adding support for the use case or improving the error or the documentation or the tests, feel free to suggest some direction and I'll have a look in my spare time. Thanks anyway!

dtolnay:
c858291a9ff62bcbd123363407cd73e05ab1a46a should work as of cxx 1.0.76.

Issue #52, opened 2020-03-02
Support passing function pointers across the FFI


dtolnay:
i.e. function parameters with the type `fn(...) -> ...`.

```rust
#[cxx::bridge]
mod ffi {
    extern "Rust" {
        fn f(arg: u8, callback: fn(u8));
    }
}

fn f(arg: u8, callback: fn(u8)) {
    callback(256 - arg);
}
```

dtolnay:
Passing function pointers from Rust to C++ was implemented in #79 + #85. It remains to implement passing the reverse direction.

wendajiang:
I think the direction from C++ to Rust, as that need C++ function definition compile together? 

What is I need is the case: 

```Rust
extern "Rust" {
  fn callback_from_cpp_definition(cb: fn(i32));
}
```
cb definition is in cpp, and when compile Rust code to lib, there is only declaration, and the lib need to ld with other cpp definition finallly. How can I get this?

wendajiang:
> @dtolnay What's the plan for solving this? You are going to use std::function? Can I help with the implementation?

Do you have any progress about this?

I want this feature, and as cxx not support it, so I have to use 
```
#[no_mangle]
extern "C" fn register_callback(extern "C" fn(c_int)) {
  todo!()
}
```
to register function pointer(with specific function signature) on a static memory, and in Rust call the `callback`

Can I do something for support this feature? @dtolnay and this design is ok?

ios-yifan:
![image](https://github.com/dtolnay/cxx/assets/22848630/d31165f5-b985-4338-ad52-a24523c4cf98)

error: passing a function pointer from C++ to Rust is not implemented yet

rustyifan:
> [Here's a pattern](https://github.com/dtolnay/cxx/compare/master...adetaylor:cxx:std-function) to use `std::function` with current cxx facilities.

I used this, but it does not support dynamic libraries,  can you give me a hand,  tks!!

Undefined symbols for architecture arm64:
            "_cn$rongcloud$cxxbridge1$MyFunction$call", referenced from:
                imlib::ffi::MyFunction::call::hb153764149e230b4 in imlib.4dg0tx688zy81d57.rcgu.o

https://github.com/ios-yifan/rust-std.git

ios-yifan:
> > [Here's a pattern](https://github.com/dtolnay/cxx/compare/master...adetaylor:cxx:std-function) to use `std::function` with current cxx facilities.
> 
[...]

PR #1100, opened 2022-09-21
Added support for std::function

paandahl:
I put this up as a draft to get feedback; the code still needs cleanup / documentation.

I have added support for `std::function`, with the following interface:

```rust
#[cxx::bridge]
pub mod ffi {
    pub struct CallbackArgs<'a>(
        u8,
        &'a Vec<u8>,
        Box<OpaqueStruct>,
    );
    
    extern "Rust" {
        type OpaqueStruct;
        fn t_take_ref_func(v: &CxxFunction<CallbackArgs, u16>);
    }
}

fn r_take_ref_func_tuple_args(f: &CxxFunction<ffi::CallbackArgs, u16>) {
    let ret = f.call(
        ffi::TestArgs(
            2,
            vec![42],
            Box::new(OpaqueStruct { ... }),
        )
    );
}
```

Usage from C++:

```c++
r_take_ref_func_tuple_args([&](
    uint8_t arg,
    rust::Vec<uint8_t> arg2,
    rust::Box<tests::OpaqueStruct> arg3
) {
    ...
    return (uint16_t)200;
});
```

## Explanation

The `CxxFunction<A, R>` takes two template arguments.

* `A` - the arguments. Can be one of:
    1. A shared or opaque Rust type that is declared in the crate
    2. A tuple struct (most internal types supported)
* `R` - the return type.

If `A` is a tuple struct, the contents will be spread out, so that each element will match a function argument. The tuple struct itself evaporates at the language boundary.

The callback can also be passed by value, wrapped in a UniquePtr.

## Rationale

I wanted the API to resemble those of CxxVector, UniquePtr, etc. The tuple struct is not ideal, but necessary to allow for multiple arguments.

I considered looking into a wrapper that would expose the callback as a Fn trait object, but that seems out of line with the rest of the API.

## Possible improvements

* The code can probably be simplified / cleaned up a bit more
* Some things are skipped for now - checking-logic, documentation, etc.

## Discussion

How was this functionality imagined? Is this the way to go, or are there better ways to solve the problem at hand?

Closes: #52

### DIFF
--- a/gen/src/cfg.rs
+++ b/gen/src/cfg.rs
@@ -113,6 +113,7 @@ impl Api {
         match self {
             Api::Include(include) => &include.cfg,
             Api::Struct(strct) => &strct.cfg,
+            Api::TupleStruct(_) => &CfgExpr::Unconditional,
             Api::Enum(enm) => &enm.cfg,
             Api::CxxType(ety) | Api::RustType(ety) => &ety.cfg,
             Api::CxxFunction(efn) | Api::RustFunction(efn) => &efn.cfg,
--- a/gen/src/namespace.rs
+++ b/gen/src/namespace.rs
@@ -8,7 +8,7 @@ impl Api {
             Api::CxxType(ety) | Api::RustType(ety) => &ety.name.namespace,
             Api::Enum(enm) => &enm.name.namespace,
             Api::Struct(strct) => &strct.name.namespace,
-            Api::Impl(_) | Api::Include(_) | Api::TypeAlias(_) => Default::default(),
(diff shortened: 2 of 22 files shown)


casimcdaniels:
This is impeccable timing as I was just trying to implement C++ callbacks called from Rust.

kdewald:
Hey all! Is there a status update on this feature?

noufelez:
Hello everyone, is this feature going to be merged. ? :) @dtolnay @paandahl 

We are looking into CXX at https://easymile.com/ and we'd love to use it and possibly contribute if there are things we can do.

Thanks!

owenshoemaker:
It has been a few months and just wanted to check if there was any update. This would be super helpful.

Issue #447, opened 2020-11-11
Distinguish static member function on shared structs


XiangpengHao:
Regard https://github.com/dtolnay/cxx/pull/437, what's your opinion on how to distinguish static member function from non-static ones?
For example:
```rust
#[cxx::bridge]
mod ffi {
    struct A {
        a: f32,
    }

    extern "Rust" {
        fn foo(self: &A) -> A; 
        fn bar(self: &A) -> A;
    }
}

impl ffi::A {
    fn foo() -> A {
        A { a: 0 }
    }

    fn bar(&self) -> A {
        A { a: self.a }
    }
}
```

In the generated header file, `foo` should be a static function while `bar` should be non-static.

dtolnay:
A Rust signature should be emitted as a non-static member function if and only if it has a `self` argument. I want to support static member functions but those shouldn't be written with a `self` argument.

Maybe like this? :

```rust
#[cxx::bridge]
mod ffi {
    struct A {
        a: f32,
    }

    extern "Rust" {
        #[Self = "A"]
        fn foo() -> A; 
        fn bar(self: &A) -> A;
    }
}
```

XiangpengHao:
I'm also thinking about this:

```rust
#[cxx::bridge]
mod ffi {
    struct A {
        a: f32,
    }

    extern "Rust" {
        #[struct = "A"]
        fn foo() -> A; 

        #[struct = "A"]
        fn bar(&self) -> A;
    }
}
```

1. `struct` sounds more intuitive than `Self` 
2. attribute on `bar` is also madantary, just to be explict and have a uniform api
3. having (2), we can define `bar(&self)` so the function signature is consistent with rust conventions.

This will break existing code though.

mikhailzagurskiy:
I suggest using another approach - using nested modules to declare static member functions:

```c++

struct SimpleClass {
    static void simpleMethod();
}
```

```rust
#[cxx::bridge]
mod ffi {

    pub mod SimpleClass {
        extern "C++" {
            fn simpleMethod();
        }
    }
}

fn main() {
    SimpleClass::simpleMethod();
}
```

I think, it was really what we want for static member functions

dtolnay:
I am on board with that!

jsgf:
My first instinct was to try:
```
#[cxx_name = "SimpleClass::simpleMethod"]
fn simplething();
```
but this fails with 
```
 1 │         #[cxx_name = "SimpleClass::simpleMethod"]
[...]

PR #1102, opened 2022-09-27
`impl` Blocks and Static Methods

maurer:
Adds support for using `extern` `impl` blocks to describe methods and static/associated functions on a type.

Requires a companion patch to `rustc`, so this may take some time to be able to land.


ahayzen-kdab:
Note that there are discussions around possible static methods API here #447, although it's not clear if there was a preferred API.

dtolnay (commented):
Equivalent functionality was implemented without new syntax in https://github.com/dtolnay/cxx/pull/1546.

Thanks anyway for this PR!

PR #1111, opened 2022-10-14
Update codespan-reporting to build with -Zminimal-versions

lopopolo:
`codespan-reporting` 0.11.0 does not build with `-Zminimal-versions` because it has a constraint
 on `termcolor = 1` whereas it depends on the `Copy` trait being implemented on types in `termcolor`
which were not added until 1.0.4.

See:

- https://github.com/brendanzab/codespan/issues/322
- https://github.com/brendanzab/codespan/commit/e99c867339a877731437e7ee6a903a3d03b5439e

### DIFF
--- a/gen/build/Cargo.toml
+++ b/gen/build/Cargo.toml
@@ -19,7 +19,7 @@ experimental-async-fn = []
 
 [dependencies]
 cc = "1.0.49"
-codespan-reporting = "0.11"
+codespan-reporting = "0.11.1"
 once_cell = "1.9"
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

lopopolo:
Apologies, but it looks like the commit I referenced from codespan-reporting never made it out into a release on crates.io. Thanks for merging this, but it does not appear to have fixed compilation with a minimal versions lockfile.

PR #1112, opened 2022-10-15
Factor out the Fail lambda from all trycatch calls

dtolnay:
This unblocks adding an overload of `operator()` so that `fail` can be called with a `std::string`, not just a `const char *`, and can skip the strlen computation in that case.


dtolnay:
- Followup: #1114

Issue #462, opened 2020-11-12
Need an explicit way for cxxbridge to produce header/source file


XiangpengHao:
The `cxxbridge-cmd` is not ideal:
1. The `cxxbridge-cmd` version might be different from `cxx` version, which can cause problems
2. It's hard to track file changes outside rust project, i.e. determine the best time to re-generate the header/source file

I used the side-effect from `cxx_build::bridge(...)` to generate the necessary files https://github.com/XiangpengHao/cxx-cmake-example/blob/27dfab58ff4888d4ede71fdad45a7d112640d2b0/rust_part/build.rs#L2 and then copy those files to desired locations https://github.com/XiangpengHao/cxx-cmake-example/blob/27dfab58ff4888d4ede71fdad45a7d112640d2b0/rust_part/CMakeLists.txt#L20

It would be great if we have some ways to explictly write the ouput file, maybe something like
```rust
// build.rs
fn main() {
    cxx_build::bridge("src/lib.rs").header("/path/lib.rs.h").source("/path/lib.rs.cpp");

    println!("cargo:rerun-if-changed=src/lib.rs");
}
```

gui81:
I agree this would really be helpful.  The build environment always has to be in sync with the version of cxx, which might encourage locking on an outdated version.

Be-ing:
I have recently overhauled cxx-qt's build process and stumbled over this problem. cxx-qt has its own code generation steps in addition to cxx. Initially, we had cxx-qt-builder generating C++ code from build.rs but not compiling it, instead relying on CMake to compile the generated C++ code. This was an ugly situation. cxx-qt-builder had to run at CMake's configure time so it could pass the paths of the generated files to CMake (involving a convoluted process of serializing C++ source code and file paths to JSON :dizzy_face: ).

[...]

Issue #949, opened 2021-10-11
Unpredictable output path for generated files


emartinfigma:
We are trying to integrate cxx into a larger CMake-based build system.

Following some other examples we have a `build.rs` that contains:

```
fn main() {
    let _ = cxx_build::bridge("src/lib.rs");
}
```

Ultimately cmake generates a command line that looks like:

```
cargo build --target-dir /.../build/emscripten-1.40.1/rust --target=wasm32-unknown-emscripten --release
```

What's super confusing here is that, when executed by Ninja (which prints the command line it's running), that ends up writing `lib.rs.cc`/`.h` under

`build/emscripten-1.40.1/rust/cxxbridge/...`

but if I copy paste the same command from the command line, it puts them under

`build/emscripten-1.40.1/rust/wasm32-unknown-emscripten/cxxbridge/...`

In both cases, they contain a directory tree like

```
├── cxxbridge
│   ├── fullscreen-link (our crate name)
│   │   └── src
│   │       ├── lib.rs.cc -> ...symlink to a file in the build output that has a hash in the name...
```

For the life of me I cannot figure out what is the difference between how these two commands are run.  It must be obeying some environment variable or something?  Could the names of the path components leading up to the input or output dir matter?  What else can I investigate to try figure out the cause of this?

dtolnay:
I don't know of anything in this crate which would behave that way. I bet it's Cargo doing it. Notice the following:

```console
$ cargo new --lib repro
     Created library `repro` package

$ cd repro
$ cargo build
   Compiling repro v0.1.0
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s

$ find target -name librepro.rlib
target/debug/librepro.rlib

$ cargo clean
$ cargo build --target=wasm32-unknown-emscripten
   Compiling repro v0.1.0
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s

$ find target -name librepro.rlib
target/wasm32-unknown-emscripten/debug/librepro.rlib
```

which is a very similar pattern.

emartinfigma:
Yes, the `--target` flag reliably puts the Rust outputs in the `wasm32-unknown-emscripten` subdir for me.  The real mystery here is how the cxx-generated outputs `lib.rs.cc` etc. seem to either be in that subdir or not for reasons I cannot gauge in my execution environment.  

[...]

Issue #229, opened 2020-07-21
Getting cxxbridge working in a cmake codebase


Nehliin:
Hello, 
I'm a bit confused about the usage of the cxxbridge tool and I'm sure I've just misunderstood how to use it. I've tried to emulate the hello world example in this repo but my company is using cmake and this would be part of a much larger codebase so I tried to avoid using build.rs scripts. Our use case would be that we want to gradually replace an existing C++ code base with Rust so I want to link a Rust lib to the C++ executable. 

First off both the header and cc file generated from my `lib.rs` file contains a definition of the shared struct which of course causes compile errors if I were to include the header in the cc file without adding ifdefs. The shared struct I've defined contains a `Vec<u8>`  but none of the generated files created or the `cxx.h` file in this repo contains a definition of the private methods such as  `drop()` in the generated C++ code. Where is that definition expected to come from?

dtolnay:
There has been some experimentation with CMake in #156; perhaps it would be useful as a starting point.

<br>

> both the header and cc file generated from my `lib.rs` file contains a definition of the shared struct which of course causes compile errors if I were to include the header in the cc file without adding ifdefs

The generated cc file is not intended to include the generated header file. Is it necessary in your use case to be able to include it?

I filed #230 to follow up with ifdefs.

<br>

> The shared struct I've defined contains a `Vec<u8>` but none of the generated files created or the `cxx.h` file in this repo contains a definition of the private methods such as `drop()` in the generated C++ code

The generated files and cxx.h are all in C++, whereas the definition for how to drop a Vec\<u8\> is something that can only be provided by Rust. The `cxx` Rust crate provides those definitions (concretely [in this file](https://github.com/dtolnay/cxx/blob/0.3.4/src/symbols/rust_vec.rs)). You would need to make sure that linking includes the Rust code as well as the C++ code.

Nehliin:
Thanks for the quick response @dtolnay I'll take a look at the cmake pr! I don't think it's necessary in my use case I'm mainly just experimenting at the moment.

Nehliin:
[...]

PR #1120, opened 2022-10-28
cxx-build: symlink headers to GENERATED_HEADER_DIR

Be-ing:
Symlink (or copy on Windows) generated headers to a directory specified by the GENERATED_HEADER_DIR environment variable, if this variable is set. This makes cxx-build usable by C++ build systems if they set GENERATED_HEADER_DIR before calling cargo and add it to their include paths. This is a better solution than the CLI tool for several reasons:

* Don't need to build/distribute/install a separate program
* Saves build time because cxx crates are not built redundantly for the CLI tool and the macro
* No risk of version mismatches between CLI tool and cxx macro crate
* One library crate that uses cxx can be usable in applications built either with Cargo or a C++ build system

Fixes https://github.com/dtolnay/cxx/issues/462

### DIFF
--- a/gen/build/src/lib.rs
+++ b/gen/build/src/lib.rs
@@ -424,6 +424,16 @@ fn generate_bridge(prj: &Project, build: &mut Build, rust_source_file: &Path) ->
     let shared_cc = prj.shared_dir.join(&prj.include_prefix).join(rel_path_cc);
     let _ = out::symlink_file(header_path, shared_h);
     let _ = out::symlink_file(implementation_path, shared_cc);
+
+    if let Ok(export_path) = env::var("GENERATED_HEADER_DIR") {
+        out::symlink_file(
+            header_path,
+            PathBuf::from(export_path)
+                .join(&prj.include_prefix)
+                .join(rel_path_h),
+        )?;
+    }
+
     Ok(())
 }
 


gen/build/src/lib.rs
```diff
@@ -179,6 +179,8 @@ impl Project {
             TargetDir::Unknown => scratch::path("cxxbridge"),
         };
 
+        println!("cargo:rerun-if-env-changed=GENERATED_HEADER_DIR");
```
dtolnay, line 182: This use of an environment variable is not compatible with Cargo's expectations around build script behavior. As written, this definitely isn't something we can land because it breaks any build script that does not already contain its own accurate "cargo:rerun-if-…" lines. If cxx-build prints this line, cargo will not rerun the build script for *any* other relevant reason, for example even source file changes in the file holding the #\[cxx::bridge\]. Meanwhile not printing this line is also not correct, because build scripts would end up not seeing changes to this environment variable.

Would you be able to propose some other ways to handle what you need in cxx-qt?

Be-ing: >  breaks any build script that does not already contain its own accurate "cargo:rerun-if-…" lines

Wouldn't such a build script kinda be broken already? The example in cxx-build's documentation clearly shows the build script should have `cargo:rerun-if-changed` lines for the bridge files and [Cargo's documentation](https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection) recommends this:

> it is recommended that every build script emit at least one of the rerun-if instructions

On the other hand, since it is already the build script's responsibility to print `rerun-if-changed` for the bridge files, it would be consistent to leave `rerun-if-env-changed=GENERATED_HEADER_DIR` to the build script...

dtolnay: It would generally not already be kinda broken. `rerun-if-changed` is optional and Cargo has the sane default behavior of rerunning if any file in the project has changed. For example just from the first page of https://crates.io/crates/cxx-build/reverse_dependencies, here are several build scripts that are not broken today but would be broken by adding a `rerun-if-env-changed` into cxx-build:

- https://github.com/MaterializeInc/rust-protobuf-native/blob/protobuf-native-v0.2.1+3.19.1/protobuf-native/build.rs
- https://github.com/tsurucapital/quadprogpp-rs/blob/v0.1.0/quadprogpp-sys/build.rs
- https://github.com/risc0/risc0/blob/v0.11.1/risc0/core/build.rs (and 4 other build.rs in that repo)

Be-ing: Thinking about this further, either this must be the build script's responsibility, or cxx-build would need to only print this for bridges with `extern Rust` sections. Crates that only have `extern C++` sections would not need this.

Be-ing: Even among crates with `extern Rust` sections, they might not be designed to be used with build systems besides Cargo. For example, the `extern Rust` could only be used to interact with C++ files within the crate that are compiled with cc by the Cargo build script. In that case, `rerun-if-changed-env=GENERATED_HEADER_DIR` would be wrong. So the decision whether to print `rerun-if-changed-env` must be left to the downstream build script.

The primary use case for this is for Rust libraries written to be linked into a specific C++ application. I doubt people will want to use this to create C++ bindings for general purpose Rust libraries, though that would technically be possible.

Be-ing:
Note that if the cxx library has hand written C++ headers, its build script is responsible for copying those to GENERATED_HEADER_DIR (alternatively, the C++ build system could add the path to the crate's source code to its include paths, but I wouldn't recommend that).

Be-ing:
PR to make use of this with CMake using Corrosion: https://github.com/corrosion-rs/corrosion/pull/247

dtolnay (changes requested):
I got a chance to skim how corrosion and CMake handle integration of Cargo-based code, and some of the repos and issues that have been linked by all the discussions on this topic, and it reinforced my past impression that the whole thing is overwhelmingly a disaster (in contrast to buck+reindeer or bazel+raze). I feel strongly that this PR would not be a net improvement to corrosion's situation, and I don't think it is sufficient or necessary for handling cxx headers, so I'll close the PR.

Some examples of things that go wrong with the setup you have envisioned:

- If CMake depends on crate A which depends on crate B which depends on crate C, where A and C involve cxx-build, then both A's and C's generated headers would end up in the GENERATED_HEADER_DIR, which is not correct. Downstream code must only be allowed to import from A's headers.

- If either crate uses [CFG.exported_header_dirs](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_dirs), [CFG.exported_header_prefixes](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_prefixes), or [CFG.exported_header_links](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_links), the content of those directories would not be reflected in GENERATED_HEADER_DIR.

The behavior and experience implemented by cxx-build are specific to Cargo builds. If corrosion would like to stick non-Cargo build steps downstream of a Cargo-based invocation of cxx-build, the only correct way to implement that is by sufficiently accurately pretending to be Cargo in the rest of the build as well. For example: create a crate containing:

```rust
// lib.rs
#[cxx::bridge]
pub mod ffi {}
```

```rust
// build.rs
fn main() {
    cxx_build::bridge("lib.rs").compile("...");
}
```

In this crate, cxx-build guarantees that exactly the right include-directories will get set up as documented in https://cxx.rs/build/cargo.html, and handed to the underlying C++ compiler invocation. On top of that, the cc crate guarantees that a `CXX` environment variable will be respected in choosing what C++ compiler to invoke. So corrosion can build this crate with a `CXX` environment variable that points to a program that parses and extracts the include-directories from the compiler command line arguments.

Be-ing:
> Downstream code must only be allowed to import from A's headers.

If crate C can use crate A's headers, why shouldn't external C++ code?

> If either crate uses [CFG.exported_header_dirs](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_dirs), [CFG.exported_header_prefixes](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_prefixes), or [CFG.exported_header_links](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_links), the content of those directories would not be reflected in GENERATED_HEADER_DIR.

Right, that would be up to build scripts to handle if they need that. Or alternatively I could put more work into this to copy/symlink headers from those paths into GENERATED_HEADER_DIR.

> Corrosion can build this crate with a CXX environment variable that points to a program that parses and extracts the include-directories from the compiler command line arguments.

That would be a very convoluted workaround.

Be-ing:
> If CMake depends on crate A which depends on crate B which depends on crate C, where A and C involve cxx-build

Oh, are you talking about a case where crate B uses crate A for bindings to some C++ library, but crate B doesn't re-export anything from crate A? In that case, you're right, a single environment variable might have unintended consequences. This could be overcome by including the crate name in the environment variable name, for example GENERATED_HEADER_DIR_my_cxx_library

Be-ing:
> If either crate uses [CFG.exported_header_dirs](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_dirs), [CFG.exported_header_prefixes](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_prefixes), or [CFG.exported_header_links](https://docs.rs/cxx-build/1.0.80/cxx_build/static.CFG.html#cfgexported_header_links), the content of those directories would not be reflected in GENERATED_HEADER_DIR.

This applies equally to the cxxbridge CLI tool.

Be-ing:
> I don't think it is sufficient or necessary for handling cxx headers, so I'll close the PR.

Please don't dismiss use cases outside of what you have personally encountered. The existing solution with cxxbridge is cumbersome, confusing (consider that all but one of the CMake examples listed in the documentation have a build.rs using cxx_build in addition to cxxbridge), and brittle. Quite a few users have communicated (#462, #949, #229) this. I'm not saying it can't work, but that getting it working requires a lot more work than it needs to and there's lots of room for the build process to go wrong.

> my work codebase using Buck has several hundred binding crates successfully built on it.

(from https://github.com/corrosion-rs/corrosion/pull/244#issuecomment-1295232222)
Clearly you have more experience with complex dependency graphs involving cxx than I do. I'm glad you've pointed out pitfalls in my code, but I don't think that means the whole approach should be dismissed. I think we can come up with a mutually agreeable solution together if the discussion isn't shut down prematurely. If https://github.com/dtolnay/cxx/pull/1120#issuecomment-1295299586 isn't sufficient, perhaps some way could be added for symlinking/copying the generated headers to be opt-in for build scripts using cxx-build.

PR #1123, opened 2022-11-03
cxx: implement slice support for nostd

twilfredo:
## OVERVIEW:
This patch adds experimental support for us to use `slices` in a nostd/freestanding environment, aim is to run c++ <=> cxx-rs <=> rust on a baremental target environment and use slices, with the intention of supporting more features later.

## CFG
In the application `Cargo.toml` file, the following feature cfgs are required:

```
[dependencies]
cxx = { ... , default-features = false }

[build-dependencies]
cxx-build = { ... , default-features = false }
```

Then build with:
```
RUSTFLAGS='--cfg cxx_experimental_no_alloc' cargo build
```
### Build Notes
and in the particular `.cpp` file you may need to declare the following

```
void __assert_func (const char *__a, int, const char *__b, const char *__c) {
	while (true) {}
}

void *__gxx_personality_v0;
```

## FUNCTIONALITY:
Currently, this only supports slices (outside of trivial features).

## TESTING:
Tested by compiling:
	- cargo test (run existing tests)
	- cxx/demo/ running it.
	- compiling with a arm toolchain setup with cmake/corrosion and running the binary on QEMU arm with basic logic assertions [[1]]( https://github.com/twilfredo/qemu_misc/tree/master/bm_arm).

Current testing has been done in the direction of C++ -> Rust with a simple callback test to C++. A
simple test setup can be seen here [[2]](https://github.com/twilfredo/qemu_misc/blob/master/bm_arm/main.cpp).

## TODO:
- Get features such as `Results<Ok, Err>` working.

- When we build for the none `std` case (no `default-features`), instead of creating a symlink to the original `cxx.h` file, we copy it over and define the macro to disable stdlib dependencies. Perhaps there's a better way to do this?

## NOTES:
By default, all the standard features are enabled gaurded by the `#ifndef CXXBRIDGE1_RUST_STD`, so this **shoudn't** break anything.

[1] https://github.com/twilfredo/qemu_misc/tree/master/bm_arm 
[2] https://github.com/twilfredo/qemu_misc/blob/master/bm_arm/main.cpp

Signed-off-by: Wilfred Mallawa <wilfred.mallawa@wdc.com>


twilfredo:
@dtolnay ping!

twilfredo:
ping!

GregoryTravis:
I would love to use this.

Issue #1124, opened 2022-11-04
ManuallyDrop generate -Wshadow warnings


hfiguiere:
The generated code for `ManuallyDrop` is as follow:
```c++
template <typename T>
union ManuallyDrop {
  T value;
  ManuallyDrop(T &&value) : value(::std::move(value)) {}
  ~ManuallyDrop() {}
};
```

`value` is both a member and the argument name of the constructor triggering a warning with `-Wshadow`.

hfiguiere:
cxx version 1.0.78

dtolnay:
Closing along with https://github.com/dtolnay/cxx/pull/1125#pullrequestreview-1171713185.

Thanks anyway for the PR!

PR #1125, opened 2022-11-04
Issue #1124 - Fix -Wshadow warning in ManuallyDrop

hfiguiere:
Just change the generated code.

### DIFF
--- a/gen/src/builtin.rs
+++ b/gen/src/builtin.rs
@@ -279,7 +279,7 @@ pub(super) fn write(out: &mut OutFile) {
         writeln!(out, "  T value;");
         writeln!(
             out,
-            "  ManuallyDrop(T &&value) : value(::std::move(value)) {{}}",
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
I think this code is fine as written.

If you are using GCC: it's a longstanding bug in GCC that `-Wshadow` gets triggered in this situation. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147. Clang's `-Wshadow` does not trigger here.

How are `-W` settings normally dealt with in C++ projects? Hopefully you have a way to build third-party code like `cxx` using different warning settings than what's applied to your own first-party code.

hfiguiere:
No it's more complicated to see "per file" than to fix that bug in cxx. cxx generated files are compiled just like any other C++ files in my project.

And the issue you link is 6 year old, still open.

This is very disappointing decision.

PR #1174, opened 2023-02-11
C++: Fix build on no-except compilers

riidefi:
This fix was necessary for my project (C++ depending on Rust) to compile. The approach here is conservative, only deferring from the existing behavior when the compiler *explicitly does not support exceptions* via feature-test macros. There may be other approaches here, like attempting to compile a test file with `throw`.


dtolnay (approved):
Thanks!

PR #1181, opened 2023-02-25
Minimal support for Rust type aliases

schreter:
The intention is to provide support to re-import Rust types already exported elsewhere into other bridges, creating equivalence of `extern "Rust"` types on the C++ side.

Currently, the support is limited to re-exporting the type under the same name and the C++ namespace must be explicitly specified. Then, C++ functions can use the same type.

### DIFF
--- a/gen/src/namespace.rs
+++ b/gen/src/namespace.rs
@@ -6,9 +6,10 @@ impl Api {
         match self {
             Api::CxxFunction(efn) | Api::RustFunction(efn) => &efn.name.namespace,
             Api::CxxType(ety) | Api::RustType(ety) => &ety.name.namespace,
+            Api::TypeAlias(ety) => &ety.name.namespace,
             Api::Enum(enm) => &enm.name.namespace,
             Api::Struct(strct) => &strct.name.namespace,
-            Api::Impl(_) | Api::Include(_) | Api::TypeAlias(_) => Default::default(),
+            Api::Impl(_) | Api::Include(_) => Default::default(),
         }
     }
 }
--- a/syntax/check.rs
+++ b/syntax/check.rs
@@ -3,7 +3,8 @@ use crate::syntax::report::Errors;
 use crate::syntax::visit::{self, Visit};
 use crate::syntax::{
     error, ident, trivial, Api, Array, Enum, ExternFn, ExternType, Impl, Lang, Lifetimes,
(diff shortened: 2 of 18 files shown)


gen/src/namespace.rs
```diff
@@ -6,9 +6,10 @@ impl Api {
         match self {
             Api::CxxFunction(efn) | Api::RustFunction(efn) => &efn.name.namespace,
             Api::CxxType(ety) | Api::RustType(ety) => &ety.name.namespace,
+            Api::TypeAlias(ety) => &ety.name.namespace,
```
wvwwvwwv, line 9: Merge the branch with the above?

syntax/check.rs
```diff
@@ -500,6 +501,25 @@ fn check_api_type_alias(cx: &mut Check, alias: &TypeAlias) {
         let msg = format!("derive({}) on extern type alias is not supported", derive);
         cx.error(derive, msg);
     }
+
+    if alias.lang == Lang::Rust {
+        let ty = &alias.ty;
+        if let RustType::Path(path) = &ty {
+            // OK, we support path
+            if let Some(last) = path.path.segments.last() {
+                if last.ident != alias.name.rust {
```
wvwwvwwv, line 510: Not sure if we really need to prevent this - some may really want to use different names for the same Rust type.

schreter: The problem is the C++ side. Effectively, what we want to do is to publish a forward declaration of the struct in question in the C++ code in the proper namespace (that's why also the namespace should match, otherwise the types are different on C++ side - currently, this is not checked, but probably should be).

If we'd like to support renaming the object, that would be possible, but again, that would require materializing both the forward declaration *and* some kind of type alias (e.g., `using` statement), so we would have both. The name would be accessible by both names and the original name taken anyway.

I.e., someone trying to rename the type to prevent name clashes would still end up with name clashes. Therefore, better to simply not allow renames.

The only way to do this properly would be to expose the original type in the original namespace and the alias name in a new namespace. However, due to the restriction on the Rust side, we only have access to tokens within our bridge, so there is no access to the original namespace (that's also the reason why the namespace must be re-specified).

gen/src/namespace.rs
schreter: Unfortunately, that's not possible, since those have different inner types.

wvwwvwwv: Acknowledged.

syntax/check.rs
wvwwvwwv: I see, thanks - I misunderstood the intention.

PR #1182, opened 2023-02-25
Fix: pick up changed tests.h/cc to rebuild the bridge

schreter:
Previously, modifying tests.h/cc would not trigger rebuild of the bridge, preventing efficient test development.


dtolnay (approved):
Thanks!

PR #1183, opened 2023-02-26
Fix: Prevent crash on OOM in error copy

schreter:
When the error message is copied, the allocation may fail. This would terminate the process, since it's called from a `noexcept` function.

Make the allocation `nothrow` and handle `nullptr` in `what()` method appropriately, returning a generic message saying that the message cannot be allocated.

This also prevents a crash if `what()` is called on a moved object (which is technically UB, but still better not to crash).

### DIFF
--- a/src/cxx.cc
+++ b/src/cxx.cc
@@ -449,8 +449,9 @@ static_assert(!std::is_same<Vec<std::uint8_t>::const_iterator,
               "Vec<T>::const_iterator != Vec<T>::iterator");
 
 static const char *errorCopy(const char *ptr, std::size_t len) {
-  char *copy = new char[len];
+  char *copy = new char[len + 1];
   std::memcpy(copy, ptr, len);
+  copy[len] = 0;
   return copy;
 }
 
--- a/src/cxx.cc
+++ b/src/cxx.cc
@@ -448,10 +448,14 @@ static_assert(!std::is_same<Vec<std::uint8_t>::const_iterator,
                             Vec<std::uint8_t>::iterator>::value,
               "Vec<T>::const_iterator != Vec<T>::iterator");
 
-static const char *errorCopy(const char *ptr, std::size_t len) {
-  char *copy = new char[len + 1];
-  std::memcpy(copy, ptr, len);
-  copy[len] = 0;
+// Copy the error message into a private buffer and return it.
+// If the allocation fails, nullptr is returned.
+static const char *errorCopy(const char *ptr, std::size_t len) noexcept {
+  char *copy = new (std::nothrow) char[len + 1];
+  if (copy) {
+    std::memcpy(copy, ptr, len);
+    copy[len] = 0;
+  }
   return copy;
 }
 
@@ -497,7 +501,12 @@ Error &Error::operator=(Error &&other) &noexcept {
   return *this;
 }
 
-const char *Error::what() const noexcept { return this->msg; }
+const char *Error::what() const noexcept {
+  if (this->msg)
+    return this->msg;
+  else
+    return "<message not allocated>";
+}
 
 namespace {
 template <typename T>


schreter:
Note: this subsumes the bugfix for invalid read https://github.com/dtolnay/cxx/pull/1179.

dtolnay (changes requested):
Terminate on OOM matches the behavior of all other code in the repo, on both the Rust side and C++ side. Is this PR at all useful to you without a whole bunch of other refactoring of all the other usages of operator new and liballoc?

dtolnay:
(For example, stuff like https://github.com/dtolnay/cxx/blob/1.0.91/src/cxx_vector.rs#L164 and https://github.com/dtolnay/cxx/blob/1.0.91/include/cxx.h#L333)

schreter:
@dtolnay 

> Terminate on OOM matches the behavior of all other code in the repo, on both the Rust side and C++ side.

That is our problem that currently, there is no clean interface how to prevent crashes. The issue is, we are integrating our Rust code in a C++ process which runs potentially a multi-terabyte process, which simply can't just fall down. Therefore, yes, we have have an issue that we have to carefully handle OOM situations and map OOM in general to a panic/exception. Fortunately, almost all Rust library code is already prepared for that, so it's not a big deal panicking/throwing an exception in the allocation (and yes, we know that it's not yet fully implemented, but we are here on the bleeding edge).

> Is this PR at all useful to you without a whole bunch of other refactoring of all the other usages of operator new and liballoc?

It's of course not optimal, it's just a small fix for this particular issue. In general, throwing in a `noexcept` method is a no-go in our project.

Containers need to be fixed, too, but we currently don't use them yet. So there was no investment there yet.

In fact, we use custom exception handlers anyway (I didn't push the change yet to upstream, though, since it needs some polishing), so we are not affected much by not having this in, I just pushed it as a warm-up for completeness (whenever I see a bug/potential issue I try to fix it right away or at least note it down for later).

IMHO, for this particular issue, it is definitely correct way to go. A comparabe example would be `std::make_exception_ptr()`. It is `noexcept` and also won't crash - it will rather give you an instance of `std::bad_alloc` inside of `std::exception_ptr` before crashing the program unnecessarily.

src/cxx.cc
```diff
@@ -448,9 +448,14 @@ static_assert(!std::is_same<Vec<std::uint8_t>::const_iterator,
                             Vec<std::uint8_t>::iterator>::value,
               "Vec<T>::const_iterator != Vec<T>::iterator");
 
-static const char *errorCopy(const char *ptr, std::size_t len) {
-  char *copy = new char[len];
-  std::memcpy(copy, ptr, len);
+// Copy the error message into a private buffer and return it.
+// If the allocation fails, nullptr is returned.
+static const char *errorCopy(const char *ptr, std::size_t len) noexcept {
+  char *copy = new (std::nothrow) char[len + 1];
```
dtolnay, line 454: I am on board with handling OOM in this function without crashing the process. However I think the implementation given in this PR is not correct.

This change makes `errorCopy` return nullptr if it fails to allocate memory to copy the error message. That means `cxxbridge1$error` will return nullptr to Rust.

https://github.com/dtolnay/cxx/blob/03b0c8d41a1ffb470a3e43bbe7cc9c4aeb6b04e3/src/cxx.cc#L458-L460

That means this Rust code will be holding a `NonNull<u8>` which is actually null, which is immediate language-level UB.

https://github.com/dtolnay/cxx/blob/03b0c8d41a1ffb470a3e43bbe7cc9c4aeb6b04e3/src/result.rs#L45-L50

Since this is UB we cannot say what really happens, but naively it's likely the null ptr ends up as the `err` field of `union Result`, which makes the union look like it contains an `ok` not `err`.

https://github.com/dtolnay/cxx/blob/03b0c8d41a1ffb470a3e43bbe7cc9c4aeb6b04e3/src/result.rs#L51-L52

https://github.com/dtolnay/cxx/blob/03b0c8d41a1ffb470a3e43bbe7cc9c4aeb6b04e3/src/result.rs#L13-L24

C++ will use the nullness of that pointer to determine whether an error has occurred in the Rust function, so it will appear that it has not.

https://github.com/dtolnay/cxx/blob/03b0c8d41a1ffb470a3e43bbe7cc9c4aeb6b04e3/gen/src/write.rs#L1126-L1128

The C++ side will continue as if a successful result has been written into the function's output location, which it has not. In reality the output location will be uninitialized. C++ will move or copy out of it which will be UB, and anything the caller does with it will be further UB, including heap corruption or other such issues depending on what data structure the return type is.

dtolnay (changes requested):

PR #1188, opened 2023-03-02
minor nit in tutorial.md

nyurik:
function comment should follow rust's fn md docs for consistency


dtolnay (commented):
I think this is all right as written. The content of this comment is aimed at the reader of the tutorial, not at the consumer of the MultiBuf type. A documentation comment on MultiBuf would be substantially different from what is written here.

PR #1194, opened 2023-03-16
[cxxbridge] Add lib.rs back in

rdelfin:
I've been using this crate within a bazel project, using [rules_rust's](https://github.com/bazelbuild/rules_rust) crate universe feature. We recently noticed that after upgrading to 1.0.92, the build no longer works. It seems the issue is that [crate_universe](https://bazelbuild.github.io/rules_rust/crate_universe.html) is unable to support binary-only crates. I was able to get this working by extracting the crate into its own repo, but that is, unsurprisingly hard to maintain (you can see my workaround in [this example](https://github.com/rdelfin/rust_cxx_build_test)). I want to raise this up to the rules rust repo, but in the meantime would you be willing to put this file back in? Thanks!


dtolnay (commented):
I would want to understand why "crate_universe is unable to support binary-only crates" and what might be done about that in crate_universe, on what timescale, before I would be okay with adopting this workaround.

PR #1196, opened 2023-03-16
Add feature flag to skip building cxx.cc

dpaoliello:
This adds a new feature flag, `skip_cxxbridge1_build` to skip building `cxx.cc` in cxx's build.rs.

# Why add this feature flag?
* To support build systems that wrap Cargo, but don't allow the use of the `cc` crate, or don't have the environment correctly setup for a C++ build.
* Allow dependents to avoid rebuilding `cxx.cc` multiple times if there are multiple Rust static libraries (.so or .lib) being built separately that are then linked together.
* To allow a single Rust library to be consumed by multiple C++ projects that are using different STL implementations.

# Naming discussion
* Prefixed with `skip` to indicate that `cxx.cc` is built by default (avoiding breaking changes for dependents using `default-features = false`).
* `cxx.cc` is built into a library called `cxxbridge1` - not sure if there is a better way to reference this? (Maybe `skip_static_library_build`, `skip_support_library_build` or `skip_runtime_build`?)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ default = ["std", "cxxbridge-flags/default"] # c++11
 "c++20" = ["cxxbridge-flags/c++20"]
 alloc = []
 std = ["alloc"]
+skip_cxxbridge1_build = []
 
 [dependencies]
 cxxbridge-macro = { version = "=1.0.92", path = "macro" }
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
I believe you can accomplish what you need using https://doc.rust-lang.org/1.68.0/cargo/reference/build-scripts.html#overriding-build-scripts, without this change.

Issue #1201, opened 2023-03-23
use-of-uninitialized-value with CxxString::push_str()


parasyte:
I get this error message from MemorySanitizer on a very simple test using only safe APIs.

```rust
#[test]
fn test_push_str() {
    let_cxx_string!(s = "");

    s.as_mut().push_str("Hello, ");
    s.as_mut().push_str("world");
    s.as_mut().push_str("!");

    assert_eq!(s.to_str(), Ok("Hello, world!"));
}
```

And the output from the test run:

```
$ MSAN_SYMBOLIZER_PATH="$(which llvm-symbolizer-15)" RUSTFLAGS='-Zsanitizer=memory -Zsanitizer-memory-track-origins' RUSTDOCFLAGS='-Zsanitizer=memory -Zsanitizer-memory-track-origins' cargo test -Zbuild-std --target x86_64-unknown-linux-gnu -- test_push_str

     Running tests/cxx_string.rs (target/x86_64-unknown-linux-gnu/debug/deps/cxx_string-541ff2a9405221a2)

running 1 test
==28682==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55a92766f06e in core::str::validations::run_utf8_validation::h822c24368d3f2331 /home/jay/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/str/validations.rs:232:23
    #1 0x55a92766f06e in core::str::converts::from_utf8::h3df7c88a528f5971 /home/jay/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/str/converts.rs:89:11
    #2 0x55a926cc09de in cxx::string::CxxString::to_str::h8fc4bc119aa82020 /home/jay/other-projects/cxx/src/cxx_string.rs:141:9
    #3 0x55a926803f21 in cxx_string::test_push_str::h0d903d066679d4aa /home/jay/other-projects/cxx/tests/cxx_string.rs:32:16
    #4 0x55a926802c18 in cxx_string::test_push_str::_$u7b$$u7b$closure$u7d$$u7d$::hdcfb002a984f9e41 /home/jay/other-projects/cxx/tests/cxx_string.rs:25:20
[...]

PR #1202, opened 2023-03-23
Implement fmt::Write for CxxString

parasyte:
This allows formatted writes (e.g. with the `write!()` macro) directly on a `CxxString`. This can eliminate a needless allocation from formatting to a `String` then converting to a `CxxString`.

While testing this, I ran into #1201. Nothing else stood out as unusual.

### DIFF
--- a/src/cxx_string.rs
+++ b/src/cxx_string.rs
@@ -5,7 +5,7 @@ use alloc::borrow::Cow;
 #[cfg(feature = "alloc")]
 use alloc::string::String;
 use core::cmp::Ordering;
-use core::fmt::{self, Debug, Display};
+use core::fmt::{self, Debug, Display, Write};
 use core::hash::{Hash, Hasher};
 use core::marker::{PhantomData, PhantomPinned};
 use core::mem::MaybeUninit;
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
I am not sure whether std::fmt::Write or std::io::Write is more appropriate. In the Rust standard library, `String` implements fmt::Write but not io::Write, while `Vec<u8>` implements io::Write and not fmt::Write. `CxxString` has similarities to both in different ways.

If we add fmt::Write now then adding io::Write later is a breaking change because it would make `cxx_string.write_fmt(…)` ambiguous in the case that you have both traits in scope.

Thoughts?

parasyte:
Great observation! I can add an impl for io::Writer on `CxxVec` to this PR. That should address the semver hazards. Are there any other concerns that come to mind?

parasyte:
@dtolnay I have addressed the concerns with breaking changes by implementing both `fmt::Write` and `io::Write`. I decided not to do anything with `CxxVector` because I don't need it and it's more work to test.

dtolnay (approved):
Thanks!

Issue #1049, opened 2022-05-09
Pass `const void*` to C++ binded function


hamptonmoore:
Hello,

I'm trying to create Rust bindings for the library for the library https://github.com/richgel999/fpng, I'm trying to bind against their function with a signature of 
```
  bool fpng_encode_image_to_memory(const void* pImage, uint32_t w, uint32_t h, uint32_t num_chans, std::vector<uint8_t>& out_buf, uint32_t flags = 0);
```
So far I have 
```
        unsafe fn fpng_encode_image_to_memory(pImage: *const i32, width: u32, height: u32, num_chans: u32, out_buf: &CxxVector<u8>, flags: u32) -> bool;
```

Which according to cxx has the signature of  `bool (*)(const int*, unsigned int, unsigned int, unsigned int, const std::vector<unsigned char>&, unsigned int)` while the intended signature would be `bool (*)(const void*, unsigned int, unsigned int, unsigned int, std::vector<unsigned char>&, unsigned int)` 

How do I inform cxx I would like to pass a void pointer to some data? I tried a lot related to passing [u8]s, but could never get it working

dtolnay:
There isn't yet a `void` binding built in, but it's easy to DIY. In your header add:

```cpp
using c_void = void;
```

and then in your `extern "C++"` module:

```rust
    type c_void;
    unsafe fn fpng_encode_image_to_memory(pImage: *const c_void, width: u32, …);
```

PR #1204, opened 2023-03-31
Add support for c_void as an opaque C++ type

dpaoliello:
Fix for #1049

Makes `c_void` a built-in primitive type that maps to an opaque C++ type 'void'.


rockwotj:
@dtolnay any opposition to this? It'd be nice to have.

SunDoge:
Is there any progress?

besquared:
+1

ctdaws:
+1, took me quite a bit of digging to figure out how to work around this. Would be great to have as a standard feature

orowith2os:
I need this, in order to pass void pointers through FFI to convert into their appropriate gobject types. Would appreciate some movement here.

PR #1214, opened 2023-05-18
cxx-gen: Add span() to Error type

LeonMatthesKDAB:
The Error type exposed in cxx-gen is very minimal currently. For use in CXX-Qt, we'd like to access the span if the Error has one. This would allow us to greatly improve our build script diagnostics (very similar to how CXX displays errors itself).

cc: @ahayzen-kdab
See: https://github.com/KDAB/cxx-qt/issues/536

### DIFF
--- a/gen/lib/src/error.rs
+++ b/gen/lib/src/error.rs
@@ -9,6 +9,16 @@ pub struct Error {
     pub(crate) err: crate::gen::Error,
 }
 
+impl Error {
+    /// Returns the span of the error, if available.
+    pub fn span(&self) -> Option<proc_macro2::Span> {
+        match &self.err {
+            crate::gen::Error::Syn(err) => Some(err.span()),
+            _ => None,
+        }
+    }
+}
+
 impl From<crate::gen::Error> for Error {
     fn from(err: crate::gen::Error) -> Self {
         Error { err }
--- a/gen/lib/src/error.rs
+++ b/gen/lib/src/error.rs
@@ -42,3 +42,33 @@ impl StdError for Error {
         self.err.source()
     }
 }
+
+impl IntoIterator for Error {
+    type Item = Error;
+    type IntoIter = IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        match self.err {
+            crate::gen::Error::Syn(err) => IntoIter::Syn(err.into_iter()),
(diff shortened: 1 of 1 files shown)


dtolnay (changes requested):
In general there are going to be multiple different spans for all the different errors, so an API that only exposes one span is not going to be adequate.

```rust
use quote::quote;

fn main() {
    let tokens = quote! {
        #[cxx::bridge]
        mod ffi {
            enum Enum {}
            unsafe extern "C++" {
                unsafe fn f(ptr: *const &usize);
                fn f(self: &Enum);
            }
        }
    };
    let opt = cxx_gen::Opt::default();
    if let Err(err) = cxx_gen::generate_header_and_cc(tokens, &opt) {
        eprintln!("{:#?}", err);
    }
}
```

```console
Syn(
    Error(
        [
            "C++ does not allow pointer to reference as a type",
            "explicit #[repr(...)] is required for enum without any variants",
            "unsupported receiver type; C++ does not allow member functions on enums",
        ],
    ),
)
```

Could you look into how the cxxbridge-cmd renders multiple errors and figure out what cxx-gen needs to add?

LeonMatthesKDAB:
Hi, thanks for the review.
Thanks for pointing this out. Wasn't expecting a syn::Error to contain multiple spans.

Will look at cxxbridge-cmd to try and figure out a more adequate API.
I'll let you now when I have something better to suggest.

dtolnay:
Mimicking syn::Error with an IntoIterator impl seems all right to me. Or `generate_header_and_cc` could just return Vec\<cxx_gen::Error\> as the error type. Feel free to do whatever is easiest for cxx-qt-build, since I do not personally use cxx-gen for anything. Or @adetaylor might have an opinion.

LeonMatthesKDAB:
@dtolnay I've implemented `IntoIterator` for cxx_gen::Error.
That works out alright for Cxx-Qt (see https://github.com/KDAB/cxx-qt/pull/547) and is compatible with previous uses of the cxx-gen API.

As mentioned in the Commit, I only implemented this on Error and not &Error, as that would require the internal cxx::gen::Error to implement Clone, so to keep the API changes minimal, I left this out.
For Cxx-Qt this means we have to expand the errors directly when creating our diagnostic type, but that's not too bad.

dtolnay (approved):
Thanks!

PR #1218, opened 2023-05-27
chore: clippy let-underscore-untyped

nyurik:
Clippy keeps complaining about
  https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_untyped

Seems like a noop, esp considering that @dtolnay has filed a bug in
  https://github.com/rust-lang/rust-clippy/issues/10411

One issue though is that MSRV in .clippy.toml is not matching the one in Cargo.toml:
* .clippy.toml:  msrv = "1.48.0"
* Cargo.toml: rust-version = "1.60"

Which one is the accurate one?  I would assume its the Cargo.toml, in which case it might be better to apply this change?

### DIFF
--- a/book/src/async.md
+++ b/book/src/async.md
@@ -60,7 +60,7 @@ pub async fn do_thing(arg: Arg) -> Ret {
 
     ffi::shim_doThing(
         arg,
-        |context, ret| { let _ = context.0.send(ret); },
+        |context, ret| { _ = context.0.send(ret); },
         context,
     );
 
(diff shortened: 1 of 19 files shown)


dtolnay (commented):
This lint was removed from clippy for being misguided, so I would prefer not to follow its guidance.

nyurik:
@dtolnay thx, i wasn't sure if its applicable or not.  What about the difference in MSRV?

dtolnay:
I think it's fine. Some of the crates in the workspace have a different supported rust-version from one another.

nyurik:
> I think it's fine. Some of the crates in the workspace have a different supported rust-version from one another.

But if they are different, shouldn't clippy rules be relevant to each specific crate?  I honestly don't understand the need for a global clippy-only MSRV -- maybe there is some other reason I'm missing?

PR #1219, opened 2023-05-27
chore: grammar & spelling

nyurik:
Minor grammatical fixes, nothing big

P.S. I will try to make more significant contribs, but these basic fixes are usually good to get a bit more acquainted with the code structure

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
         # Our Windows and macOS jobs are the longest running, so exclude the
         # relatively slow compiletest from them to speed up end-to-end CI time,
         # except during cron builds when no human is presumably waiting on the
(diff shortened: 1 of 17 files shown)


dtolnay (commented):
I took the descriminant->discriminant fix in f5bcbece4a8d19df0d7f0fdab548c78fb0d6f603.

The typo in third-party/bazel/defs.bzl needs to be fixed in https://github.com/bazelbuild/rules_rust/pull/1981 because that is generated code. Changing it here will not eliminate the typo.

The rest of these, I think are fine either way, so I would prefer to keep as originally written.

Thanks!

nyurik:
@dtolnay there is one more -- https://github.com/dtolnay/cxx/compare/master...nyurik:cxx:grammar?expand=1

nyurik:
Oh wow, I was not even aware that panick is a valid spelling,  nvm

Issue #1225, opened 2023-06-14
MSVC warning C4365: signed/unsigned mismatch in cxx.cc operator<< overloads


wravery:
After a recent compiler upgrade on Windows, I started seeing a warning that there's a signed/unsigned mismatch in `cxx.cc`:

```cpp
std::ostream &operator<<(std::ostream &os, const String &s) {
  os.write(s.data(), s.size());
  return os;
}
```

The `s.size()` method returns `std::size_t`, but `os.write` takes a signed [`std::streamsize`](https://en.cppreference.com/w/cpp/io/streamsize)

I think the same probably applies to the equivalent `Str` operator below that, but in my case the warning was treated as an error, and it seems to have stopped compilation rather than reporting that one.

PR #1226, opened 2023-06-15
Explicit cast from std::size_t to std::streamsize in cxx.cc

wravery:
This is a suggested fix for #1225.


wravery:
When I ran `cargo test` locally with `rustc` 1.70, some of the `stderr` messages were a little different. I rebuilt with `TRYBUILD=overwrite` and confirmed that the output has changed a little, but the actual errors reported are the same. If the CI build runs into the same mismatch, I'll push another commit with those changes.

dtolnay (approved):
Thanks!

PR #1233, opened 2023-06-30
tests: ensure that extern "Rust" methods on C++ types work

ahayzen-kdab:
This allows for developers and crates that are generating both C++ and Rust code to have a C++ method implemented in Rust without having to use a free method and passing through the C++ "this" as an argument.


ahayzen-kdab:
While trying to simplify our free functions and wrapper methods going between generated C++ classes, Rust structs, and methods implemented in a C++ context in Rust. We found that it was possible to implement a Rust method onto a C++ type. As CXX only generates the implementation for it and we are generating the C++ class we can specify the right header match. Then this avoids us having to have a free function to proxy C++ calls to the Rust calls as we can then go directly between.

As this was slightly non-obvious we wanted to submit a test to ensure that the behaviour doesn't regress and to ensure that it is accepted behaviour that we can depend on.

tests/cxx_gen.rs
```diff
@@ -32,3 +32,48 @@ fn test_impl_annotation() {
...
+            fn rust_method_cpp_context(self: Pin<&mut CppType>);
+        }
+    }
+"#;
+
+// Ensure that implementing a Rust method on a C++ type only causes generation
+// of the implementation.
+//
+// The header should be implemented in the C++ class definition and the Rust
+// implementation in the usual way.
```
dtolnay, line 53: I don't understand what this middle paragraph means.

I will make a followup PR to reword this.

dtolnay (approved):
Thank you!

tests/cxx_gen.rs
dtolnay: #1571

PR #1240, opened 2023-07-06
Remove usage of std::cerr

### LINKED ISSUE #1239: Avoid using c++'s `stream`?
As in here:

https://github.com/dtolnay/cxx/blob/8308e991189ce6d8c2eb535a5369772d97ff8fb5/src/cxx.cc#L79

we use `std::cerr` which is known to escalate the binary size and is unsuitable for embedded environments. Many people disable exceptions for smaller binary sizes, so I think this request is valid. Can we use `fprintf` instead?

I can submit a PR if accepted.

yujincheng08:
The usage of std::cerr significantly increases binary size. However, many people disable exceptions for binary size. Also, std::terminate also uses std::cerr to print an abort message and then abort. Here we manually print the abort message and thus use std::abort instead.

Fix #1239

### DIFF
--- a/src/cxx.cc
+++ b/src/cxx.cc
@@ -1,4 +1,5 @@
 #include "../include/cxx.h"
+#include <cstdio>
 #include <cstring>
 #include <iostream>
 #include <memory>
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #1243, opened 2023-07-07
Add feature `cxxbridge` to allow user manually compile it

yujincheng08:
Sometimes we want to manually compile the `cxxbridge` (e.g., we use our own toolchain for embedded devices). Thus we introduce a `cxxbridge` which is enabled by default and users can disable it and generate and compile their own bridges using customized toolchains.

I saw https://github.com/dtolnay/cxx/pull/1196#pullrequestreview-1344892828 but I still think this feature is necessary. For example, if I have multiple static libraries and need to integrate some Rust codes into them.  If I am not able to disable compelling `cxxbridge`, I will have multiple `cxxbridge` on different static libraries. And linking will fail with duplicate symbols. If I could disable `cxxbridge`, I can only compile it on a commonly linked library.

In this patch, I also export two functions (i.e., `export_cxx_bridge` and `generate_header_and_cc_with_path`) for c++ build systems (like cmake or gnumake). There are many reasons that we still need a c++ build system (e.g., cargo does not support post-build scripts). With the two newly exported functions, we can integrate rust codes into c++ build systems.


yujincheng08:
@dtolnay sorry for pinging. any comment?

dtolnay (commented):
I am skeptical about this PR because I have not seen any other library that compiles native code in its build script need to do this.

The use case you described is multiple static libraries which each include the cxx crate. I am not clear on how eliding the C++ half of the cxx crate from those static libraries addresses the use case. All the static libraries would still include the Rust half of the cxx crate and those would be duplicate symbols.

yujincheng08:
Here you are a project (I am one of its maintainers):

https://github.com/topjohnwu/Magisk/tree/master/native/src

It now needs to use a modified version of cxx to avoid building cxxbridge.


> All the static libraries would still include the Rust half of the cxx crate and those would be duplicate symbols.

I think it won't include duplicate symbols. The rust part (and all deps) wont be included in the output static libraries. They are stored in separate libraries which will link only into the final elf. However, cxxbridge will include in our own static libraries and thus cause duplicate symbol errors.

PR #1251, opened 2023-08-05
Relativize symlinks where possible

### LINKED ISSUE #1250: Emit with relative symlinks?
I'm using https://crane.dev to build a Cargo project with Nix. Crane has an elegant feature where it can precompile all dependencies in one build step (which is cached) so unless you change your dependencies, no recompile of dependencies is required. This drastically speeds up builds.

The approximate process by which it does this is:

```
cargo build
mv target $out
```

Unfortunately, cxx seems unique in our tech stack in that it creates absolute symlinks in its output, which end up like:

```
lrwxr-xr-x  1 root  wheel  189  1 Jan  1970 ./release/build/mycrate-461e0524ef218d12/out/cxxbridge/include/mycrate/src/lib.rs -> /private/tmp/nix-build-cargo-package-deps-0.0.1.drv-1/source/target/release/build/mycrate-461e0524ef218d12/out/cxxbridge/include/mycrate/src/lib.rs.h
```

Unfortunately, the tempdir is immediately deleted. I'm wondering if there's a reason other than simplicity that absolute symlinks are used here, and whether a PR might be accepted that adjusts them to being relative? At the moment I am resolving this problem by running `symlinks -rc $PWD` on the directory after the build has occurred, but this feels more fragile than fixing the source.

I _think_ that this would also have the effect of meaning that cxx builds are not broken after a move of the project directory, but am not 100% on how cache invalidation works.

j-baker:
Attempt to relativize symlinks. Fixes #1250.

- Do not relativize where target or link are relative paths.
- Do not relativize where CARGO_TARGET_DIR is set, as this indicates that relativizing might not be helpful.
- Do not relativize where source or target contain `..`

It feels a bit weird where it is because the module is filesystem-specific, cxx-agnostic, but I justified it to myself on the basis that it's _cxx's_ interface for interacting with the filesystem specifically for the purpose of writing output.

### DIFF
--- a/gen/build/src/out.rs
+++ b/gen/build/src/out.rs
@@ -1,8 +1,8 @@
 use crate::error::{Error, Result};
 use crate::gen::fs;
 use crate::paths;
-use std::io;
-use std::path::Path;
+use std::path::{Path, PathBuf};
+use std::{env, io};
 
 pub(crate) fn write(path: impl AsRef<Path>, content: &[u8]) -> Result<()> {
     let path = path.as_ref();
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #1257, opened 2023-08-13
Add bazel annotations as Cargo.toml package metadata

dtolnay:
Adapted from the "@using_cxx" example added by https://github.com/bazelbuild/rules_rust/pull/2103.

I plan to send a rules_rust PR to read this package metadata and obviate the `annotations = { "cxx": [crate.annotation( .... )] }` for crates that provide it.

I considered a more TOML-native form for `additive_build_file_content`, but just using multiline string literal containing Starlark seems all right. Support for the structured form can be introduced separately.

```toml
[package.metadata.bazel]
extra_aliased_targets = { cxx_cc = "cxx_cc" }
gen_build_script = false

[[package.metadata.bazel.additive_build_file_content]]
[package.metadata.bazel.additive_build_file_content.cc_library]
name = "cxx_cc"
srcs = ["src/cxx.cc"]
hdrs = ["include/cxx.h"]
include_prefix = "rust"
includes = ["include"]
linkstatic = true
strip_include_prefix = "include"
visibility = ["//visibility:public"]
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,22 @@ members = ["demo", "flags", "gen/build", "gen/cmd", "gen/lib", "macro", "tests/f
 targets = ["x86_64-unknown-linux-gnu"]
 rustdoc-args = ["--cfg", "doc_cfg", "--generate-link-to-definition"]
 
+[package.metadata.bazel]
+additive_build_file_content = """
+cc_library(
+    name = "cxx_cc",
+    srcs = ["src/cxx.cc"],
+    hdrs = ["include/cxx.h"],
+    include_prefix = "rust",
+    includes = ["include"],
+    linkstatic = True,
+    strip_include_prefix = "include",
+    visibility = ["//visibility:public"],
+)
+"""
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
https://github.com/bazelbuild/rules_rust/pull/2124

illicitonion:
Did this PR intend to also add `:cxx_cc` to the `deps` attribute of `cxx`?

When I upgraded past it, I started needing to manually add that annotation, or my binaries wouldn't link because they were missing symbols defined in this library... Is there something I'm meant to do in my calling code to add the dependency?

dtolnay:
Good call, we should include that. For whatever reason the "using_cxx" example from https://github.com/bazelbuild/rules_rust/pull/2103 works without it. It looks like they chose to put the `:cxx_cc` dependency somewhere else other than as a dependency of the `cxx` Rust target.

Want to send a PR for rules_rust and this repo? A rules_rust change will be needed because in https://github.com/bazelbuild/rules_rust/pull/2124 I failed to handle `deps`.

illicitonion:
Will do - thanks for the clarification!

PR #1262, opened 2023-08-25
Add CxxVector::new for creating an empty vector

capickett:
When constructing a CxxVector to pass to C++, you have to end up writing your own utility function that constructs a `UniquePtr<CxxVector<...>>` and add that to your projects cxx bridge. Creating an empty vector seems like a simple enough task to afford having an API for it on `CxxVector`.


dtolnay (changes requested):
Thanks, looks good so far. You'll also need some changes in gen/src/write.rs (this is why linking failed with MSVC in CI).

dtolnay (approved):
Thanks!

dtolnay:
Published in cxx 1.0.107.

PR #1274, opened 2023-10-03
cxx_vector: add API for reserving capacity and retrieving it

jhonboy121:


russelltg:
Curious, why was this abandoned? I just saw this after filing almost the exact same PR (#1300)

PR #1275, opened 2023-10-09
Use full path relative to $CARGO_MANIFEST_DIR when compiling

illicitonion:
Bazel by default builds in $CARGO_MANIFEST_DIR, but can be instructed to build elsewhere (e.g. so that toolchains which involve relative paths can use correct relative paths).

This allows that to function.


dtolnay (approved):
Thanks!

PR #1300, opened 2024-01-04
CxxVector: implement reserve() and capacity()

russelltg:
Efficient insertions into CxxVector requires reserving additional capacity. This also adds a way to query current capacity.

Also implement `extend`, which can use the size hint


dtolnay (approved):
Thanks!

PR #1302, opened 2024-01-05
Include printing of "cargo:rerun-if-changed" in tutorial

dennisschagt:
I ran into some compilation issues while going through the tutorial step-by-step.
Looking at the tutorial files in the repo, I expect this was caused by me missing the `cargo:rerun-if-changed` lines in the build script (without them, changes to the .cc and .h files did not have any effect).
I'm adding those on the tutorial page so other people will be aware of them from the start.


dtolnay (approved):
Thanks!

PR #1305, opened 2024-01-10
Add custom deleter support to UniquePtr

darkwisebear:
Add the possibility to use shared trivial types as deleters for objects managed by std::unique_ptr / UniquePtr.

### DIFF
--- a/src/unique_ptr.rs
+++ b/src/unique_ptr.rs
@@ -13,6 +13,7 @@ use cxx::type_id;
 
 /// Rust representation of the default deleter of std::unique_ptr, `std::default_delete<T>`
 #[derive(Default, Copy, Clone, Debug)]
+#[repr(C)]
 pub struct DefaultDeleter([u8; 0]);
 
 unsafe impl ExternType for DefaultDeleter {
--- a/src/unique_ptr.rs
+++ b/src/unique_ptr.rs
@@ -20,6 +20,22 @@ unsafe impl ExternType for DefaultDeleter {
     type Kind = Trivial;
 }
 
+#[cfg(not(target_os = "macos"))]
+#[repr(C)]
+struct UniquePtrInner<T, D> {
+    deleter: D,
+    repr: MaybeUninit<*mut c_void>,
+    ty: PhantomData<T>,
+}
+
+#[cfg(target_os = "macos")]
+#[repr(C)]
+struct UniquePtrInner<T, D> {
+    repr: MaybeUninit<*mut c_void>,
+    deleter: D,
+    ty: PhantomData<T>,
+}
+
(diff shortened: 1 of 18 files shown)


darkwisebear:
@dtolnay Does this PR have chances to be merged as-is or is there still something I need to do with it in order to get it merged?

darkwisebear:
Well, now this guy has a merge conflict. It shouldn't be a big deal to fix it but is it really worth it? Was it laying around for such a long time because the change is bad? If yes, please tell me.

On a more general side, is it worth that I also add allocator support to vector and shared_ptr? Probably not because I feel that for some reason there is no chance of such changes to ever getting merged. If that is due to lack of time I'd be glad to help here. There's a lot of pending PRs that would be embraced by the community and maybe a lot of stuff that people would be willing to contribute I guess.

src/unique_ptr.rs
```diff
@@ -9,61 +9,119 @@ use core::marker::PhantomData;
...
 }
 
-impl<T> UniquePtr<T>
+#[cfg(target_os = "macos")]
+#[repr(C)]
+struct UniquePtrInner<T, D> {
+    repr: MaybeUninit<*mut c_void>,
+    deleter: D,
+    ty: PhantomData<T>,
+}
```
dtolnay, line 38: This is not something I am willing to assume about the internal layout of std::unique_ptr.

Can you suggest some other way to implement this feature?

dtolnay (changes requested):
Thanks for the PR!

dtolnay:
I am pretty confident that this general approach is not a good fit for this library. Depending on exactly how you were planning to use this, I can recommend some alternatives:

1. Use a shared ptr that just always has one reference and is moved like a unique ptr. Shared pointers have a type-erased deleter so the Rust `SharedPtr` binding already works with custom deleters.

1. Use `using UniqueC = std::unique_ptr<C, PmrDeleterForC>` in C++ and write your own Rust `struct UniqueC` and `unsafe impl ExternType for UniqueC`, making whatever layout assumptions you consider appropriate for the set of standard libraries and platforms that your software is targeting.

1. Use `using UniqueC = std::unique_ptr<C, PmrDeleterForC>` and wire up Bindgen to produce the Rust struct for it with the correct layout for the user's platform, adding a manual `ExternType` impl on the Bindgen-generated struct to make it usable in the bridge module by value.

1. Use `using UniqueC = std::unique_ptr<C, PmrDeleterForC>` and then use `UniqueC` as an opaque C++ type in the bridge module, i.e. double boxing.

1. Make your own type-erased unique pointer struct with a well-defined layout. Wrap it in a Rust binding that attaches type information using `PhantomData`.

PR #1316, opened 2024-02-15
Implement TryInto<String> for &CxxString

tolvanea:
Added standard conversion Trait from CxxString to rust String. It is implemented only for reference of CxxString, because CxxString does usually exists in owned form on rust side .

### DIFF
--- a/src/cxx_string.rs
+++ b/src/cxx_string.rs
@@ -264,6 +264,16 @@ impl fmt::Write for Pin<&mut CxxString> {
     }
 }
 
+#[cfg(feature = "alloc")]
+#[cfg_attr(doc_cfg, doc(cfg(feature = "alloc")))]
+impl TryInto<String> for &CxxString {
+    type Error = Utf8Error;
+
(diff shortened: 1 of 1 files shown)


dtolnay (changes requested):
Thanks for the PR!

Could you share some examples of code where having this impl would be useful? I am having a hard time predicting how this is intended to be used.

Ordinarily I would expect:

```rust
let s = cxxstr.to_str()?;
s.to_owned()
```

or:

```rust
cxxstr.to_str().map(str::to_owned)
```

to be at least as good as going through TryFrom, in fact usually better by exhibiting better type inference.

tolvanea:
The idea was to make conversion between CxxString and rust String more fluent with only one call `cxxstr.into()?`. But now that I think this again, it is too far fetch to bake two distinct conversion steps (reference->owned and cxx->rust) into one.
Also, apparently `std::ffi::CString` does not even implement `TryInto<String>`, so not seeing a need for this either.

My apologies, it felt so natural at the time. Closing this.

PR #1321, opened 2024-03-03
Documentation: Use `builder.std` rather than explicit compiler flags

dcoles:
Small documentation update after running into this issue using the MSVC `cl` compiler on Windows.

Thanks for the awesome library.

---

The `-std=c++11` flag only works with GNU-style compiler frontends for MSVC this should be `/std:c++11`.
Even better is to use the builder's explicit C/C++ standard setter.


dtolnay (approved):
Thank you! I didn't know about this method.

PR #1328, opened 2024-03-21
Support for Enums

dorezyuk:
PR adds `std::variant` like type with the same memory layout as Rust's enum and codegen code to generate binding to c++ from Rust. 

The PR is quite massive and I've know that there is currently little active development at this project. Nonetheless, if the development is ever picked up, it would be great to include the feature.

It should probably also go behind a feature flag..


dphaldes:
@dorezyuk what's the status on this ?

dorezyuk:
I think its in a zombie state - I've reached out when I was working on it to @dtolnay to check how to add this into the repo (its a massive PR after all) and David said he has capacity right now to review (and maintains) such large features. 

But it does not mean that the people interested in this (especially the cpp/Rust gurus here) should not review it. Maybe this will even help to ease the burden from David

Ryex:
Well I'm not sure I qualify as a guru but I've read though the the diff and cross referenced the various docs both official and those you linked.

To the best of my knowledge you've satisfied the invariant necessary and the static asserts seem to cover all the bases.
The only think I'm a *bit* iffy on is the std::variant conversion from a valueless state.

There seems to be a few extraneous formatting changes mixed in but those might just be the un-resesolved conflicts.
This *does* need to be behind a feature flag for the c++17 requirement though.

Very much hoping this gets picked up. I don't *really* want to have to maintain a fork for a project in the simi near future.

Ryex:
@dorezyuk What are your thoughts on this being extracted into it's own macro and generator crate pair?

What I'm thinking is that there is a proc macro like cxx-qt's  that marks marks a module with a suitable enum and re-writes it to an processed enum and it's `ExternType` impl followed by `include!(concat!(env!("OUT_DIR"), "cxxenumvariant/src/path/to/file.rs/<enum>_bridge.rs"))` 

Then a paired generator crate that has a generator which is fed files with the macro like how the bridge builder is. The generator then emits a new rust file at the include point with the generated `#[cxx::bridge]` module that has the rest of the tokens form the bridge passed through alongside a block like:
```rust
  extern "C++" {
    type #enum_ident = supper:: #enum_ident
  }
```
the generator also emits a `.cc` file with the 
```cpp
class MyEnum: public ::rust::variant<Types> {
  // ...
}
```
and any headers needed

the generator run in the `build.rs` returns the paths to the generated cpp files and bridge modal stub which the user passes along to the `cxx_build` builder.

PR #1334, opened 2024-04-03
gen: allow for cfg_evaluator to be set in cxx_gen

ahayzen-kdab:
This allows for users of cxx_gen to choose a cfg_evaluator, otherwise they cannot have cfg attributes in bridges.

### DIFF
--- a/gen/lib/src/lib.rs
+++ b/gen/lib/src/lib.rs
@@ -47,7 +47,7 @@ mod syntax;
 
 pub use crate::error::Error;
 pub use crate::gen::include::{Include, HEADER};
-pub use crate::gen::{GeneratedCode, Opt};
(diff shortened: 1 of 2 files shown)


ahayzen-kdab:
When using `cxx_gen::generate_header_and_cc` https://docs.rs/cxx-gen/latest/cxx_gen/fn.generate_header_and_cc.html we can supply an `Opt` but the `cfg_evaluator` is not public and it defaults to `UnsupportedCfgEvaluator`.

This means that when using just `cxx_gen` you cannot support `#[cfg(...)]` in CXX bridges, by making this field public we can then set our own evaluator and support cfg attributes in bridges.

dtolnay (approved):
Thanks!

dtolnay:
Published in cxx-gen 0.7.121.

Issue #1317, opened 2024-02-15
Disable exceptions without setting additional env flags


cschlosser:
Hi,

I'm building on a system with `-fno-exceptions`.
The only way I found to prevent the `throw` statement from leading to a compiler error is defining `RUST_CXX_NO_EXCEPTIONS` in the `CXXFLAGS` env var as it's read by `cc`.
I don't want to set this as a global define for all builds using the environment variable if possible.

Is this the preferred solution or am I missing something?

If this is currently the preferred way of dealing with this would you be open to exposing this (and maybe exception handling in the crate in general) through a feature flag?
Or would you prefer a more implicit solution like #1174?

PR #1343, opened 2024-04-22
Add `no_exceptions` feature flag

andreiltd:
What it says on the tin.

Closes: https://github.com/dtolnay/cxx/issues/1317


ahayzen-kdab:
Note now that the `cc` crate sets `-fno-exceptions` on all WASM builds `cxx` now fails to build when using eg emsdk. https://github.com/rust-lang/cc-rs/commit/e61f4f6c92af948859e988ccec0fb9441a01976e

This might become more important, either need a way to read from the env vars to setup the define or maybe a feature like this is the way.

DavidPeicho:
Would be amazing to have this merged :)

dtolnay (commented):
Closing in favor of #1174.

Thanks for the PR!

PR #1344, opened 2024-05-07
Resolve unexpected_cfgs warning

dtolnay:
Context: https://blog.rust-lang.org/2024/05/06/check-cfg.html

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,6 @@
+#![allow(unknown_lints)]
+#![allow(unexpected_cfgs)]
+
 use std::env;
 use std::path::{Path, PathBuf};
 use std::process::Command;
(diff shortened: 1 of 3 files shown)


Urgau:
Heads up, with the release of rust-lang/cargo#13913 (in nightly-2024-05-19), there is no longer any need to allow the lint to avoid a build.rs as done in this PR. Cargo has now gain the ability to declare `--check-cfg` args directly inside the `[lints]` table with [`[lints.rust.unexpected_cfgs.check-cfg]`](https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html#check-cfg-in-lintsrust-table)[^1]:

`Cargo.toml`:
```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(trybuild)'] }
```

> Note that the diagnostic output of the lint has been updated to suggest the `[lints]` approach first. You can use it to guide you through the `--check-cfg` arguments that may need to be added.

[^1]: take effect on Rust 1.80 (current nightly), is ignored on Rust 1.79 (current beta), and produce an unused warning below

PR #1345, opened 2024-05-09
Switch to bundled buck2 prelude

dtolnay:
https://buck2.build/docs/users/advanced/external_cells/

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -4,6 +4,9 @@ prelude = tools/buck/prelude
 toolchains = tools/buck/toolchains
 none = none
 
+[external_cells]
+prelude = bundled
+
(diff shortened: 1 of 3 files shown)


dtolnay:
@JakobDegen, this is working on Linux and macOS, but not Windows.

```console
BUILD FAILED
From load at implicit location

Caused by:
    0: non-forward relative bundled path (internal error)
    1: expected a normalized path but got an un-normalized path instead: `abi\BUCK.v2`
```

dtolnay:
Fixed by https://github.com/facebook/buck2/commit/9de4fc1729c49eb7b000be7a0a5cd8d2aa6db136. Thank you!

PR #1353, opened 2024-06-05
Give char the if_unique treatment too to handle possible collisions with [u]int8_t definitions.

luqmana:
On illumos at least you have `std::is_same<char, int8_t> = true` which means a whole bunch of duplicate definition errors.


dtolnay (approved):
Thanks!

PR #1358, opened 2024-07-02
Allow going from `UniquePtr<T>` to `Pin<UniquePtr<T>>`

tbu-:
This is safe since there cannot be any other references to the interior `T` and because the `Deref`/`DerefMut` implementations are sane.

### DIFF
--- a/src/unique_ptr.rs
+++ b/src/unique_ptr.rs
@@ -84,6 +84,14 @@ where
         }
     }
 
+    /// Converts a `UniquePtr<T>` into a `Pin<UniquePtr<T>>`.
+    ///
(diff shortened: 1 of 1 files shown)


dtolnay (changes requested):
For a `T` which is not `Unpin`, cxx's `UniquePtr<T>` already enforces that safe code cannot move `T`. Can you explain more about the use case for needing `Pin<UniquePtr<T>>`? It doesn't have any invariant that is different from `UniquePtr<T>`.

tbu-:
You're correct. I didn't see that the `pin_mut` function already lets me go from a `UniquePtr<T>` to a `Pin<&mut T>`. This PR is unneeded. Sorry and thanks for the quick answer.

PR #1367, opened 2024-08-13
Ergonomics: allow constructing `rust::Slice` from any C++ container.

anforowicz:
After this commit, it is possible to explicitly construct `rust::Slice` from a reference to any continguous C++ container (any container that exposes `data` and `size` accessors).

The new constructor results in a slightly more ergonomic code, by removing the need to explicit extract and pass `c.data()` and `c.size()` at a callsite of a `rust::Slice` constructor.  The callsites using the new constructor are also more obviously correct, because they doesn't require double-checking that the passed `data` and `size` match.

The implementation of the new constructor mimics `std::span` from C++20, but for C++11 compatibility reimplements `std::size` / `std::ranges::size` in a pedestrian way (same for `std::data` / `std::ranges::data`).

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -176,6 +176,9 @@ public:
   Slice() noexcept;
   Slice(T *, std::size_t count) noexcept;
 
+  template <typename C>
+  explicit Slice(C& c) : Slice(c.data(), c.size()) {}
+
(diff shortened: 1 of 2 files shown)


dtolnay:
Yes &mdash; feel free to send a PR.

JonathonReinhart:
See #1449.

PR #1368, opened 2024-08-14
`impl<T> Read for UniquePtr<T> where ... Pin<&a mut T> : Read`.

anforowicz:
This commit implements forwarding of `Read` trait implementation from `UniquePtr<T>` to the pointee type.  This is quite similar to how `Box<T>` also forwards - see
https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-Read-for-Box%3CR%3E

Just as with `Box<T>`, the `impl` cannot be provided in the crate introducing `T`, because this violates the orphan rule.  This means that before this commit a wrapper newtype would be required to work around the orphan rule - e.g.:

    ```
    struct UniquePtrOfReadTrait(cxx::UniquePtr<ffi::ReadTrait>);
    impl Read for UniquePtrOfReadTrait { … }
    ```

After this commit, one can provide an `impl` that works more directly with the C++ type `T` (the FFI will typically require passing `self: Pin<&mut ffi::ReadTrait>`):

    ```
    impl<'a> Read for Pin<&'a mut ffi::ReadTrait> { … }
    ```

For a more specific motivating example, please see: https://docs.google.com/document/d/1EPn1Ss-hfOC6Ki_B5CC6GA_UFnY3TmoDLCP2HjP7bms

### DIFF
--- a/src/unique_ptr.rs
+++ b/src/unique_ptr.rs
@@ -10,6 +10,9 @@ use core::mem::{self, MaybeUninit};
 use core::ops::{Deref, DerefMut};
 use core::pin::Pin;
 
+#[cfg(feature = "std")]
+use std::io::Read;
+
 /// Binding to C++ `std::unique_ptr<T, std::default_delete<T>>`.
 #[repr(C)]
 pub struct UniquePtr<T>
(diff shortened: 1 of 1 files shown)


anforowicz:
Can you PTAL?

Some notes:

* I wasn't sure if additional tests were required here.  It seems to me that if the new `impl` compiles, then any additional tests wouldn't really provide extra value (maybe other than verifying that this really works end-to-end;  OTOH I've manually done this kind of verification by checking that http://review.skia.org/889876 works fine).
* It may make sense to forward other `std` traits that `Box<T>` forwards, for example: `Write`, `PartialEq`, etc.
    - Please shout if you'd like me to:
        - Cover some additional traits in this PR (which traits?)
        - Cover some additional traits in follow-up PRs (which traits?)
    - I note that not all trait forwarding of `Box<T>` applies to `cxx::UniquePtr<T>`.  For example:
        - `From<T>` takes `T` by value which is not possible for `UniquePtrTarget`s (well, maybe except `CxxString` and `CxxVector`)

dtolnay (approved):
Thanks!

I am open to separate PRs for forwarding other traits (`Write`, `PartialEq`) as someone comes across a real use case for those.

PR #1376, opened 2024-10-01
Add `_Enum!` macros to replace `_` arms when matching shared enums

nicolas-guichard:
As the docs currently state about shared enums:
> Pattern matching with match still works but will require you to write wildcard arms to handle the situation of an enum value that is not one of the listed variants.

This is unfortunate as there is no way to check that all listed enum variants are actually handled. IOW this code does not even warn:
```
mod ffi {
    enum Enum {
        A,
        B,
        C = 5,
    }
}

match t {
    Enum::A => {},
    Enum::B => {},
    _ => {},
}
```

This generates an extra `_Enum!` macro that matches all the ranges that are _not_ listed by the enum. When using that macro instead of the `_` catch-all, we get compile-time errors if we forget to match variants:
```
match t {
    Enum::A => {},
    Enum::B => {},
    _Enum!() => {},
}
// Error Enum{ repr: 5 } case not covered
```

The generated macro looks like:
```
macro_rules! _Enum {
    () => { Enum{ repr: 4..5 | 6.. } };
    ($i:ident) => { Enum { repr: $i @ 4..5 | 6.. } };
}
```

It can also be used to ergonomically retrieve the inner value in the unlisted case:
```
match t {
    Enum::A => {},
    Enum::B => {},
    Enum::C => {},
    _Enum!(value) => panic!("Unexpected Enum value {value}"),
}
```

I would have preferred something like `Enum::unlisted!`, but macro definitions are not supported in `impl`s. I think `_Enum` carries the meaning of `_` specific  to `Enum`, but I'm open to other naming/namespacing ideas.

### DIFF
--- a/syntax/discriminant.rs
+++ b/syntax/discriminant.rs
@@ -1,4 +1,7 @@
-use crate::syntax::Atom::{self, *};
+use crate::syntax::{
+    Atom::{self, *},
+    EnumRepr,
+};
 use proc_macro2::{Literal, Span, TokenStream};
 use quote::ToTokens;
 use std::cmp::Ordering;
@@ -179,7 +182,7 @@ impl Discriminant {
         }
     }
 
-    #[cfg(feature = "experimental-enum-variants-from-header")]
+    #[allow(dead_code)] // only used by cxxbridge-macro, not cxx-build
     pub(crate) const fn checked_succ(self) -> Option<Self> {
         match self.sign {
             Sign::Negative => {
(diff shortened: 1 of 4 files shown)


syntax/discriminant.rs
```diff
@@ -333,3 +336,16 @@ const LIMITS: [Limits; 8] = [
...
+
+impl EnumRepr {
+    #[allow(dead_code)] // only used by cxxbridge-macro, not cxx-build
+    pub fn limits(&self) -> Limits {
+        match self {
+            &EnumRepr::Native { atom, repr_type: _ } => {
+                Limits::of(atom).expect("Unexpected EnumRepr")
+            }
+            #[cfg(feature = "experimental-enum-variants-from-header")]
+            &EnumRepr::Foreign { rust_type: _ } => todo!(),
```
nicolas-guichard, line 348: NB: this TODO should obviously be fixed before this can be merged.

dtolnay (commented):
Thanks for the PR! This is a neat idea but I would prefer not to provide this in this library.

I would suggest making a separate proc macro that users could put on an ffi module to generate these macros for its enums.

nicolas-guichard:
For the record, I've published https://crates.io/crates/cxx-enum-gaps to do just that.

PR #1379, opened 2024-10-10
Support for parsing modules with inner attributes

makai410:
Fixes: #1378 .

### DIFF
--- a/gen/src/file.rs
+++ b/gen/src/file.rs
@@ -21,7 +21,9 @@ fn parse(input: ParseStream, modules: &mut Vec<Module>) -> Result<()> {
     while !input.is_empty() {
         let mut cxx_bridge = false;
         let mut namespace = Namespace::ROOT;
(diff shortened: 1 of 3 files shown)


gen/src/file.rs
```diff
@@ -21,7 +21,9 @@ fn parse(input: ParseStream, modules: &mut Vec<Module>) -> Result<()> {
     while !input.is_empty() {
         let mut cxx_bridge = false;
         let mut namespace = Namespace::ROOT;
+        let inner_attrs = input.call(Attribute::parse_inner)?;
         let mut attrs = input.call(Attribute::parse_outer)?;
+        attrs.extend(inner_attrs);
```
dtolnay, line 26: This is not correct. This is treating `mod outer { #![attr] mod inner {} }` as if the attr is an attribute on `inner`, instead of an attribute on `outer`.

dtolnay (changes requested):

PR #1396, opened 2024-11-12
Delete meson build

dtolnay:
I might be interested to give Meson another try in a couple years. For now, I do not get the impression that Meson's Rust support is mature enough to be worth maintaining in this repo, nor does Meson overall have any compelling advantage over a Buck2 build for my uses.


zeenix:
> nor does Meson overall have any compelling advantage over a Buck2 build for my uses.

Last I checked, Buck2 doesn't even try to compete in the same space as Meson (Linux systems programming) and hence misses essential features that Meson users take for granted (e.g installation of auxiliary data in the right paths, uninstall commands etc). These features are also important for Linux distribution packagers. There is a very good reason nobody ever even considers Buck2 in this space. Systemd, various Linux system services, GNOME and (afaik) KDE projects all use Meson (if they've ported away from autotools, which most of them have).

> For now, I do not get the impression that Meson's Rust support is mature enough

Unfortunately Meson maintainer is not fond of Cargo (and he does have some good points there) and it's been years since it even got to the stage we're at now. However, moves like this further alienates the Linux systems programming crowd and makes Rust an even harder sell than it already is.

WavyEbuilder:
Is there any criteria you're looking for regarding having upstream meson support in cxx? It's a shame to see meson go, and I'm wondering if there is anything specific that would bee needed or wanted to see meson support added upstream again. Personally, I'd go down the route of trying to not use build.rs with meson (as if upstream was to give meson support I don't really see the reasoning as to using build.rs, it's a bit messy using a Cargo build concept with... any other build system) - as noted above, a lot of Linux desktop projects use meson so it'd be something useful to have.

Not sure if this is worth mentioning, but I intend to redo the meson support (and maintain a meson.build outside of upstream, without attempting to use the build.rs and Cargo build support provided by the repo) for cxx myself at some point, so I thought it might be best to first see upstream's position for any changes in meson itself that might be desired before proceeding.

Thanks

dtolnay:
> I don't really see the reasoning as to using build.rs, it's a bit messy using a Cargo build concept with... any other build system

On the contrary, it is not messy with Buck or Bazel. From what I have seen, it is messy (I would use a worse word) with Meson only. A compiled executable that generates additional build flags for a different compiler invocation is a core competency for a build system that Meson has deliberately abdicated -- I recall there was an issue or PR that was rejected as being opposed to Meson's principles. I think someone motivated could happily use Rust with Meson, but only without crates.io packages. For cxx there are crates.io packages which are necessary.

WavyEbuilder:
Alright, understood. I appreciate your time, thanks again.

Issue #1399, opened 2024-11-15
Trick to get bazel to link libstdc++ or libc++ in a rust_binary with cxx?


sthornington:
Hi folks - I'm going to continue banging away on this, but in our bazel build, I have gotten as far as our rust_binary linking simply missing the std c++ library (either libstdc++ for a gcc toolchain or libc++ for a clang toolchain).  I feel as though this should have happened magically somehow, by virtue of linking against cc_library targets for my C++ code and the cxx-generated bindings of the ffi module, but it doesn't seem to be happening.  I have confirmed that adding e.g. `-lc++` to final link line if clang does fix the problem.  

What am I missing?  I am following the example in rules_rust/examples/crate_universe/using_cxx here: https://github.com/bazelbuild/rules_rust/blob/main/examples/crate_universe/using_cxx/BUILD.bazel

sthornington:
(noticing now that that experimental flag breaks rust linking of other stuff :( .. going to try adding `rustc_flags` to the specific `rust_binary` that is using cxx, but I am curious how other people are doing this.)

sthornington:
Hmm the problem seems isolated to compiling on clang on linux, getting this now:

```
/usr/bin/ld.gold: error: bazel-out/k8-dbg/bin/external/crate_index__cxx-1.0.130/libcxx_cc.a(cxx.o): requires dynamic R_X86_64_32 reloc against '__gxx_personality_v0' which may overflow at runtime; recompile with -fPIC
```
Bizarre....

sthornington:
Currently trying to figure out how to patch the `Cargo.toml` of `cxx 1.0.130` so that it adds a `-fPIC` to the compilation of `cxx_cc` but I can't seem to get it to stick.  I wonder if bazel is somehow scooping up the metadata file before `crate_universe` can even patch the download ..???

sthornington:
Patching this via an annotation in crates_universe seems to be impossible (double-checking in https://github.com/bazelbuild/rules_rust/issues/3004), but when I hand-edited my cargo-bazel-lock file and updated the checksum so that it was accepted, I was able to test that adding `-fPIC` as the compiler suggests fixes the issue.

PR #1400, opened 2024-11-15
add -fPIC for cxx_cc so that it can link in clang properly

sthornington:
Fixes linking in clang toolchains, e.g. https://github.com/dtolnay/cxx/issues/1399

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ cc_library(
     include_prefix = "rust",
     includes = ["include"],
     linkstatic = True,
+    copts = ["-fPIC"],
     strip_include_prefix = "include",
     visibility = ["//visibility:public"],
 )


dtolnay (changes requested):
This looks odd to me. `-fPIC` is something that needs to be seen by every C++ compiler invocation in a whole build. It isn't something that individual `cc_library` targets should be deciding to pass or not pass. Do you have more context about why this is the right fix?

sthornington:
> This looks odd to me. `-fPIC` is something that needs to be seen by every C++ compiler invocation in a whole build. It isn't something that individual `cc_library` targets should be deciding to pass or not pass. Do you have more context about why this is the right fix?

Ultimately, I was just following clang's suggestion during trying to get it all working with clang in https://github.com/dtolnay/cxx/issues/1399

```
/usr/bin/ld.gold: error: bazel-out/k8-dbg/bin/external/crate_index__cxx-1.0.130/libcxx_cc.a(cxx.o): requires dynamic R_X86_64_32 reloc against '__gxx_personality_v0' which may overflow at runtime; recompile with -fPIC
```

sthornington:
I have great stake in how this is done, but this is the only thing that fixed it for me when building with a clang cc toolchain - have you been able to build it all in Linux with clang some other way?

sthornington:
Sorry I meant "I have *no* great stake in how it all gets done", I just need to be able to link rust and C++ and cxx bindings in both gcc and clang in bazel.

sthornington:
Ok I just added it to my bazelrc and that does the same trick.  I guess I don't need this merged after all, but it's here if you want it.

sthornington:
That said, I have other issues... when compiling with `-fsanitize` stuff, it complains that that particular archive is missing __UBSAN symbols etc.

PR #1403, opened 2024-11-18
Avoid further trycatch and throw when exceptions are disabled

ahayzen-kdab:
For WASM targets the `cc` crate changed to always set `-fno-exceptions`,
before only `-fignore-exceptions` was being set elsewhere, so now programs
fail to build when using `cxx` if any `throw` or `try ... catch` are used.

> error: cannot use 'try' with exceptions disabled

Instead for trycatch just call `func()`, if the inner code has thrown it will instead abort.

> error: cannot use 'throw' with exceptions disabled

Instead of throwing just print the error and abort, any previous code that was trying to catch can't exist anyway.

Any suggestions on better routes are welcome :-)

### DIFF
--- a/book/src/binding/result.md
+++ b/book/src/binding/result.md
@@ -137,10 +137,16 @@ you'd like for the Rust error to have.
 ...namespace behavior {
 ...
 template <typename Try, typename Fail>
-static void trycatch(Try &&func, Fail &&fail) noexcept try {
+static void trycatch(Try &&func, Fail &&fail) noexcept {
+#if defined(RUST_CXX_NO_EXCEPTIONS)
   func();
-} catch (const std::exception &e) {
-  fail(e.what());
+#else
+  try {
+    func();
+  } catch (const std::exception &e) {
+    fail(e.what());
+  }
+#endif
 }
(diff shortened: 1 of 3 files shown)


ahayzen-kdab:
A further problem is that CXX itself fails in it's build script to build the crate unless RUST_CXX_NO_EXCEPTIONS is set, this only works if i set `CXXFLAGS="-DRUST_CXX_NO_EXCEPTIONS` as an env var, but maybe the build script could detect the arch in the same way as `cc` and set the define itself :thinking:

ahayzen-kdab:
Also found an alternate route through this downstream by ensuring we don't use/generate anything such as Result that would generate exception code when building for WASM.

So this pull request becomes more a question of, when `RUST_CXX_NO_EXCEPTIONS` is enabled should `trycatch` and any uses of `Result` have exception code in them which fails to build (therefore hinting to the developer that that code needs changing) or should they use alternate abort routes (therefore building and potentially silently working until they abort) ?

Which makes me wonder that these changes are not relevant. therefore i'll mark as draft and wait for feedback.

PR #1405, opened 2024-11-22
`impl<T> Write for UniquePtr<T> where ... Pin<&a mut T> : Write`.

anforowicz:
This commit implements forwarding of `Write` trait implementation from `UniquePtr<T>` to the pointee type.  This is quite similar to how `Box<T>` also forwards - see
https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-Write-for-Box%3CW%3E

This commit has quite similar, orphan-rule-related motivation as the earlier https://github.com/dtolnay/cxx/pull/1368 which covered the `Read` trait.  For a more specific motivating example, please see http://review.skia.org/c/skia/+/923337/3/experimental/rust_png/ffi/FFI.rs#254

### DIFF
--- a/src/unique_ptr.rs
+++ b/src/unique_ptr.rs
@@ -14,7 +14,7 @@ use core::mem::{self, MaybeUninit};
 use core::ops::{Deref, DerefMut};
 use core::pin::Pin;
 #[cfg(feature = "std")]
-use std::io::{self, Read};
+use std::io::{self, Read, Write};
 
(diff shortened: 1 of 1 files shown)


anforowicz:
Can you PTAL?  FWIW the earlier explanation and discussion from https://github.com/dtolnay/cxx/pull/1368 also apply here - e.g. the earlier PR says why there are no tests, and why it's okay not to worry for now about forwarding other traits.

dtolnay (approved):
Thanks!

PR #1409, opened 2024-12-01
Move cxxbridge-cmd from dependencies to build-dependencies

dtolnay:
I don't think this makes any difference for generating a lockfile or Cargo feature unification or building anything, since the dependency is not enabled on any target platform. The only intended effect is cosmetically to move cxxbridge-cmd from "Dependencies" to "Build-Dependencies" in https://crates.io/crates/cxx/1.0.132/dependencies, as it is never something that would be built for the target platform, only the host platform.


cgwalters:
FYI fallout from this in https://github.com/coreos/cargo-vendor-filterer/issues/111

PR #1415, opened 2024-12-16
Suppress `unused_unsafe` warnings within the generated code.

anforowicz:
Sometimes Rust functions declared inside the `#[cxx::bridge]` have to be marked as unsafe (e.g. if they need to use an explicit lifetime). When the actual, wrapped function is _not_ unsafe, then the generated code will result in an `unused_unsafe` warning.  For example:

```
   Compiling cxx-test-suite v0.0.0 (/usr/local/google/home/lukasza/src/github/cxx/tests/ffi)
error: unnecessary `unsafe` block
   --> tests/ffi/lib.rs:276:19
    |
276 |         unsafe fn r_return_str_via_out_param<'a>(shared: &'a Shared, out_param: &mut &'a str);
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^ unnecessary `unsafe` block
    |
note: the lint level is defined here
   --> tests/ffi/lib.rs:20:9
    |
20  | #![deny(warnings)]  // Check that expansion of `cxx::bridge` doesn't trigger warnings.
    |         ^^^^^^^^
    = note: `#[deny(unused_unsafe)]` implied by `#[deny(warnings)]`
```

The warning above comes from the following expansion:

```
unsafe fn __r_return_str_via_out_param<'a>(...) {
    // The `unsafe` block below is unnecessary if the actual, wrapped
    // function is *not* `unsafe`.
    unsafe { super::r_return_str_via_out_param(...) }
}
```

This commit avoids the warning by including an explicit `#[allow(unused_unsafe)]` in the generated code.

### DIFF
--- a/macro/src/expand.rs
+++ b/macro/src/expand.rs
@@ -1231,11 +1231,14 @@ fn expand_rust_function_shim_super(
     };
 
     let mut body = quote_spanned!(span=> #call(#(#vars,)*));
+    let mut allow_unused_unsafe = quote!();
     if unsafety.is_some() {
-        body = quote_spanned!(span=> unsafe { #body });
+        body = quote_spanned!(span=>unsafe { #body });
+        allow_unused_unsafe = quote_spanned!(span=> #[allow(unused_unsafe)]);
     }
 
     quote_spanned! {span=>
+        #allow_unused_unsafe
         #unsafety fn #local_name #generics(#(#all_args,)*) #ret {
             #body
         }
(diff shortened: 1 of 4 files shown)


anforowicz:
This PR is motivated by the desire to address the TODO at https://skia-review.googlesource.com/c/skia/+/928799/4/experimental/rust_png/ffi/FFI.rs

anforowicz:
Hmmm... when re-reading the commit / PR description, the 2nd part seems out-of-place (since there is no `foo` in the commit - this expansion comes from a separate, ad-hoc, library crate that I used to understand better what is going on).

Let me do a quick force-push to update the commit descriptions... sorry...

dtolnay (approved):
Thanks!

dtolnay:
Published in 1.0.135.

PR #1416, opened 2024-12-24
Add support for Bazel common attributes [1] in `rust_cxx_bridge`.

wmatthews-google:
This enables support for `visibility`, `testonly`, etc.

[1] https://bazel.build/reference/be/common-definitions#common-attributes


dtolnay (approved):
Thanks!

PR #1420, opened 2025-01-16
build-sys: Drop cxxbridge-cmd out of build-dependencies

cgwalters:
xref https://github.com/coreos/cargo-vendor-filterer/issues/111

This was part of an attempt to avoid lockfile skew (which has bit me) but right now cargo says this is actually an invalid dependency:

```
$ cargo metadata --format-version=1
warning: cxx v1.0.136 (/var/home/walters/src/github/dtolnay/cxx) ignoring invalid dependency `cxxbridge-cmd` which is missing a lib target
```

Yet the problem appears to be that other times cargo doesn't ignore it.

### DIFF
--- a/gen/cmd/Cargo.toml
+++ b/gen/cmd/Cargo.toml
@@ -16,6 +16,9 @@ rust-version = "1.73"
 name = "cxxbridge"
 path = "src/main.rs"
 
+[lib]
+path = "src/lib.rs"
+
 [features]
 # incomplete features that are not covered by a compatibility guarantee:
 experimental-async-fn = []
--- a/gen/cmd/src/lib.rs
+++ b/gen/cmd/src/lib.rs
@@ -1,0 +1,2 @@
+//! This is an empty stub in order to satisfy cargo's dependency
+//! resolver. This project should not be used as a library.


cgwalters:
Although maybe we could add a dummy/stub lib target to that?

cgwalters:
Or maybe reverting https://github.com/dtolnay/cxx/pull/1409/commits/58cd415524112457c238aefe5907917116bd136e is the right thing?

dtolnay (changes requested):
That warning by itself is not indicative of a problem to me. I would like to better understand how this ends up leading to the failure described in your link. Are you able to share any insight about that? What part of what `cargo metadata` is emitting to you is causing anything to break?

cgwalters:
> What part of what cargo metadata is emitting to you is causing anything to break?

The goal of cargo-vendor-filterer is to allow people who e.g. only want to build software for Linux to omit vendoring e.g. Windows dependencies. It uses `cargo metadata --filter-platform=x86_64-unknown-linux-gnu` (etc) to do that.

It seems that `cargo metadata` omits this from the list as invalid, but `cargo build` still tries to load it - and will fail, because the way `cargo vendor-filterer` works is to replace each omitted package with an empty stub.

dtolnay:
I guess I am confused what the difference is between this "platform-specific" dependency and any other platform-specific dependency. Like if we replaced the 'cfg(any())' on this line with 'cfg(windows)' then `cargo vendor-filterer` would handle it correctly? Surely not.

dtolnay:
I reproduced https://github.com/coreos/cargo-vendor-filterer/issues/111 by running `cargo vendor-filterer` on the following crate:

```toml
[dependencies]
cxx = "1"

[package.metadata.vendor-filter]
platforms = ["x86_64-unknown-linux-gnu"]
```

and I observe that it's generating these two stub crates (among others):

```toml
# vendor/windows-sys/Cargo.toml

[lib]
doctest = false
name = "windows_sys"
path = "src/lib.rs"
test = false

[package]
autobenches = false
autobins = false
autoexamples = false
autotests = false
```

```toml
# vendor/cxxbridge-cmd/Cargo.toml

[package]
autobenches = false
autobins = false
autoexamples = false
autolib = false
autotests = false
```

That seems like something to fix in `cargo-vendor-filterer`. As part of generating the empty stub src/lib.rs, it could make sure to put a `[lib]` entry in the vendored manifest if there isn't one.

The vendor/cxxbridge-cmd/Cargo.toml vendored instead by `cargo vendor` does correctly contain a `[[bin]]`. But `cargo-vendor-filterer` delegates to `cargo vendor` right? (Based on what I see in the command's output.)

Can you please look into why the `[[bin]]` written by `cargo vendor` is disappearing? Is `cargo-vendor-filterer` reading and rewriting these manifests incorrectly?

cgwalters:
> Can you please look into why the [[bin]] written by cargo vendor is disappearing? Is cargo-vendor-filterer reading and rewriting these manifests incorrectly?

This said, I guess we could change the logic for generating the stub crate to be: - If the project has a `[lib]` section, then make that a stub. If it has a `[bin]`, then make that the stub. That would not be hard, and would be a targeted fix for this.

Still though, the more I think about this the more I feel like this is at heart a cargo bug.
Either what cxx is doing right now is valid (and hence cargo-metadata should stop warning, and it should include the package *or* `cargo build` should error out too).

dtolnay:
I prefer for the issue to be fixed in cargo-vendor-filterer.

You'll need that for artifact dependencies anyway. https://doc.rust-lang.org/1.84.0/cargo/reference/unstable.html#artifact-dependencies

cgwalters:
OK. But do you have an opinion on whether there's a cargo bug here?

A different angle to look at this from is: what does it *mean* to have a `build-dependency` on a `bin` crate? It doesn't look like cargo actually compiles it.

Oh...now I see, you used this `[target.'cfg(any())'.build-dependencies]` trick to make `cargo build` in cxx ignore it - if I move it directly under `[build-dependencies]` I get the same warning from `cargo build`.

But surely this is effectively working around a cargo bug?

> You'll need that for artifact dependencies anyway. https://doc.rust-lang.org/1.84.0/cargo/reference/unstable.html#artifact-dependencies

Hum, this one is interesting. So in theory maybe what we want here is some sort of "synthetic only" dependency to basically express "**if** this crate is in the build graph, it must match this version constraint" (but if not, ignore it) right?

```
[dependencies]
cxxbridge-cmd = { version = "=1.0", artifact = "version-match-only" }
```

?

dtolnay:
Cargo's behavior is fine in my opinion. I don't think adding a new kind of dependency, distinct from an unconditionally-disabled platform-specific dependency, would be justified. And once the artifact dependencies functionality is stable we'll switch to that.

PR #1422, opened 2025-01-19
chore: update rules_cc

wep21:
- update rules_cc from 0.1.0 to 0.0.17 because 0.1.0 is yanked version. https://github.com/bazelbuild/bazel-central-registry/pull/3582


dtolnay (approved):
Thanks!

PR #1431, opened 2025-02-05
feat: add `CxxString::as_c_str() -> &CStr`

nyurik:
Exposes `std::string::c_str()` call as a `as_c_str -> &CStr` wrapper, available since 1.64


dtolnay:
Published in 1.0.138.

nyurik:
thanks for such a quick turnaround @dtolnay !

PR #1432, opened 2025-02-05
Tweak `rust::Slice` to become a C++20 `contiguous_range`.

anforowicz:
The main missing piece was that
https://en.cppreference.com/w/cpp/iterator/random_access_iterator requires that "`(a + n)` is equal to `(n + a)`", but `Slice::iterator` only supported `a + n` and didn't support `n + a`.

### DIFF
--- a/book/src/binding/slice.md
+++ b/book/src/binding/slice.md
@@ -50,7 +50,11 @@ public:
 ...template <typename T>
 ...class Slice<T>::iterator final {
 ...public:
+...#if __cplusplus >= 202002L
+...  using iterator_category = std::contiguous_iterator_tag;
+...#else
 ...  using iterator_category = std::random_access_iterator_tag;
+...#endif
 ...  using value_type = T;
 ...  using pointer = T *;
 ...  using reference = T &;
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -216,7 +216,11 @@ private:
 template <typename T>
 class Slice<T>::iterator final {
 public:
+#if __cplusplus >= 202002L
+  using iterator_category = std::contiguous_iterator_tag;
+#else
   using iterator_category = std::random_access_iterator_tag;
+#endif
   using value_type = T;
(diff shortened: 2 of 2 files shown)


anforowicz:
PTAL?

Notes:

* Motivation: I learned recently that before this commit `base::span<T>` couldn’t be easily constructed out of `rust::Slice<T>`, because `base::span`’s constructor in Chromium requires the input to be a `contiguous_range` (see [here](https://source.chromium.org/chromium/chromium/src/+/main:base/containers/span.h;l=363-373;drc=55c8af489f804ef748bf672550054786ce3fc09f) and [here](https://source.chromium.org/chromium/chromium/src/+/main:base/containers/span.h;l=993-1012;drc=55c8af489f804ef748bf672550054786ce3fc09f)).
* I wasn’t sure where to add `static_assert`s (in `cxx.h` vs in `tests/ffi/[tests.cc](http://tests.cc/)` vs somewhere else).  I think having them in `cxx.h` is useful because it also serves as documentation.  (Saying that `Slice` is a `contiguous_range` felt like a relatively low-level detail so I didn’t add that to `book/src/binding/slice.md`.)
* IIUC `cargo test` builds things in C++11 mode (?), so as-is there are no regression tests within `cxx`… :-/.  But hopefully the new `static_assert`s will make it easy to detect and diagnose regressions if they get imported into C++20-based clients (like Chromium).  So maybe this is good enough…
    - Ideally CI would test things with C++11, C++17, C++20, etc.  I may try to do this as a follow-up, but I feel that I don’t know what I am doing, so let me try that in a separate PR.  I tried making some progress here: https://github.com/anforowicz/cxx/tree/cxx20-coverage

dtolnay (approved):
Thank you!

Yes I would appreciate a follow-up PR with CI. What you linked looks good so far. I'd be happy to have a distinct job for every C++ standard since C++11, even if there's no current intentional difference between some of them in this library.

PR #1437, opened 2025-02-09
Cover C++20 in CI (fixing `char8_t`-related problem there).

anforowicz:
This is the CI coverage of C++20 (and other C++ versions) that was discussed in https://github.com/dtolnay/cxx/pull/1432.  Let me start as a "draft" PR, to quickly check if the new CI steps actually pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
       matrix:
         rust: [nightly, beta, stable, 1.82.0, 1.77.0, 1.74.0, 1.73.0]
         os: [ubuntu]
+        flags: ['']
         include:
           - name: Cargo on macOS
             rust: nightly
@@ -31,6 +32,18 @@ jobs:
             rust: nightly-x86_64-pc-windows-msvc
             os: windows
             flags: /EHsc
+          - name: C++14
+            rust: nightly
+            os: ubuntu
+            flags: -std=c++14
+          - name: C++17
+            rust: nightly
+            os: ubuntu
+            flags: -std=c++17
+          - name: C++20
+            rust: nightly
+            os: ubuntu
+            flags: -std=c++20
     env:
       CXXFLAGS: ${{matrix.flags}}
       RUSTFLAGS: --cfg deny_warnings -Dwarnings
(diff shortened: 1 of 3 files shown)


anforowicz:
Okay, I think this is ready for a review - can you PTAL?

(I see that it would have detected a regression that I tried introducing in https://github.com/dtolnay/cxx/pull/1438 as an ad-hoc test.)

.github/workflows/ci.yml
```diff
@@ -31,6 +31,22 @@ jobs:
...
             os: windows
             flags: /EHsc
+          - name: C++11
+            rust: nightly
+            os: ubuntu
+            flags: -std=c++11
+          - name: C++14
+            rust: nightly
+            os: ubuntu
+            flags: -std=c++14
```
dtolnay, line 41: Unless you add `flags` as a matrix dimension outside of `include`, this job "overwrites" the previous one, and is overwritten by the next one, ultimately leaving only the C++20 job.

```diff
         os: [ubuntu]
+        flags: ['']
         include:
```

See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations.

This is why you see only C++20 running on the PR.

Screenshot from 2025-02-09 08-50-14

dtolnay (changes requested):

.github/workflows/ci.yml
anforowicz: Thanks for catching this.  Done.

anforowicz:
Oh, but I guess now I need to check why C++11 fails...

anforowicz:
Hmmm... I guess C++11 is just not supported at this point - e.g.:

```
 cargo:warning=src/blobstore.cc: In member function 'org::blobstore::BlobMetadata org::blobstore::BlobstoreClient::metadata(uint64_t) const':
  cargo:warning=src/blobstore.cc:61:23: error: use of 'auto' in lambda parameter declaration only available with '-std=c++14' or '-std=gnu++14'
  cargo:warning=   61 |                   [&](auto &t) { metadata.tags.emplace_back(t); });
  cargo:warning=      |                       ^~~~
```

dtolnay (approved):
Thanks!

PR #1438, opened 2025-02-09
Just an ad-hoc test to see if CI really covers C++20 (DO NOT SUBMIT).

anforowicz:
This PR is not intended to be merged/landed.  I'll close it as complete once I confirm that CI catches the C++20 failure/regression that is being introduced in this PR.


anforowicz:
Abandoning - this PR was never meant to be merged/landed...

PR #1446, opened 2025-02-25
add `PartialEq` trait and its implementation to `UniquePtr`

enricobottazzi:


src/unique_ptr.rs
```diff
@@ -205,6 +205,19 @@ where
...
+impl<T> PartialEq for UniquePtr<T>
+where
+    T: PartialEq + UniquePtrTarget,
+{
+    fn eq(&self, other: &Self) -> bool {
+        match (self.as_ref(), other.as_ref()) {
+            (None, None) => true,
+            (Some(this), Some(other)) => this == other,
+            _ => false,
+        }
```
dtolnay, line 217: Simpler implementation:
```suggestion
        self.as_ref() == other.as_ref()
```

dtolnay (changes requested):
Thanks!

src/unique_ptr.rs
enricobottazzi: Updated, thanks for the suggestion

dtolnay (approved):

PR #1449, opened 2025-02-27
Add template deduction guide for Slice constructor

JonathonReinhart:
PR #1367 added an explicit Slice constructor which allows constructing from any C++ container.

This adds a C++17 template deduction guide for that constructor which allows the Slice type to be inferred from the container's contained type.

So instead of this:
```cpp
void Foo(std::span<const uint8_t> data) {
  foo(rust::Slice<const uint8_t>(data));
}
```

One could write this, omitting the `rust::Slice` template arguments:
```cpp
void Foo(std::span<const uint8_t> data) {
  foo(rust::Slice(data));
}
```

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -220,6 +220,13 @@ private:
   std::array<std::uintptr_t, 2> repr;
 };
 
+// Slice<T> template deduction guides
+#ifdef __cpp_deduction_guides
+template <typename C>
+explicit Slice(C &c)
+    -> Slice<std::remove_reference_t<decltype(*std::declval<C>().data())>>;
+#endif // __cpp_deduction_guides
+
 template <typename T>
 class Slice<T>::iterator final {
 public:
--- a/tests/ffi/tests.cc
+++ b/tests/ffi/tests.cc
@@ -1,10 +1,14 @@
 #include "tests/ffi/tests.h"
 #include "tests/ffi/lib.rs.h"
+#include <array>
 #include <cstdlib>
 #include <cstring>
 #include <iterator>
 #include <memory>
 #include <numeric>
+#ifdef __cpp_lib_span
+#include <span>
+#endif // __cpp_lib_span
 #include <stdexcept>
 #include <string>
 #include <tuple>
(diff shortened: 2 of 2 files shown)


JonathonReinhart:
FYI @anforowicz

include/cxx.h
```diff
@@ -220,6 +220,11 @@ class Slice final
   std::array<std::uintptr_t, 2> repr;
 };
 
+#ifdef __cpp_deduction_guides
+template <typename T, template <typename> typename C>
+explicit Slice(C<T> &c) -> Slice<T>;
+#endif // __cpp_deduction_guides
```
anforowicz, line 226: I note that `base::span<T>` deduction guides in Chromium use a slightly more complicated approach.  OTOH Chromium's deduction guides are built on top of C++20 `std::ranges` so I am not sure if they can be adopted here.  See:
* `base::span<T>` deduction guides from a "contiguous range" (which AFAIU is how a container => span construction works): https://source.chromium.org/chromium/chromium/src/+/main:base/containers/span.h;l=1394-1397;drc=1c8da0f65608e8e24934ce78c5f53f2357d30dfa
* `base::span<T>` constructor from a "contiguous range" (same note as in the bullet above): https://source.chromium.org/chromium/chromium/src/+/main:base/containers/span.h;l=998-1017;drc=1c8da0f65608e8e24934ce78c5f53f2357d30dfa

/cc @zetafunction who knows much more about template deduction guides than I do, and maybe can provide some guidance (pun intended? :-P) here (note that `cxx` CI covers C++14, C++17, and C++20).

PS. I don't consider myself a C++ expert, so please take my feedback with a heap of salt.

JonathonReinhart: Thanks for the feedback!

> PS. I don't consider myself a C++ expert, so please take my feedback with a heap of salt.

Well I was blissfully unaware of _template template parameters_ until I wrote this, so you're in good company. :smile: 

> Chromium's deduction guides are built on top of C++20 `std::ranges` so I am not sure if they can be adopted here.

Thanks for those `base::span<T>` references. I see they're using constraints (`requires`) which are also C++20.

I think the big difference between these is:
* You referenced a `base::span<T>` implicit constructors, so the constraints are important to avoid invalid substitutions. (I guess C++20 constraints are related to or replace SFINAE?)
* Here, we're dealing with an `explicit` constructor, so these constraints aren't as important (the programmer explicitly asked for it).

At least that's how I look at it.

tests/ffi/tests.cc
```diff
@@ -897,6 +897,12 @@ extern "C" const char *cxx_run_test() noexcept {
   ASSERT(slice_of_cpp_vec.size() == cpp_vec.size());
   ASSERT(slice_of_cpp_vec[0] == 1);
 
+#ifdef __cpp_deduction_guides
+  auto auto_slice_of_cpp_vec = rust::Slice(cpp_vec);
```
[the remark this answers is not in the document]
JonathonReinhart: Thanks for the suggestion. This helped catch a problem with `const` (moving to the other thread).

include/cxx.h
JonathonReinhart: You're right. Before seeing your comment I added a test for `const std::vector<int>` which failed.

I ended up adding a second guide which explicitly handles `const`.

But I like your idea and I'll give it a try.

JonathonReinhart: That worked perfectly, and even avoids the need for a _template  template parameter_.

JonathonReinhart: Can you explain why `std::remove_reference_t` is necessary? 

I tried without that, and the tests did try to create `Slice<int&>`. But I don't understand where the reference is coming from, since I would expect `*std::declval<C>().data()` to just be a `T` (since the signature is `T* data()`).


```
warning: cxx-test-suite@0.0.0: In file included from /usr/local/google/home/jrreinhart/code/cxx/target/debug/build/cxx-test-suite-409adf2eee31b672/out/cxxbridge/crate/tests/ffi/tests.h:2,                                                    
warning: cxx-test-suite@0.0.0:                  from tests.cc:1:                                                                                                                                                                               
warning: cxx-test-suite@0.0.0: /usr/local/google/home/jrreinhart/code/cxx/target/debug/build/cxx-test-suite-409adf2eee31b672/out/cxxbridge/include/rust/cxx.h: In instantiation of 'class rust::cxxbridge1::Slice<int&>':                      
warning: cxx-test-suite@0.0.0: tests.cc:908:53:   required from here                                                                                                                                                                           
warning: cxx-test-suite@0.0.0:   908 |     auto auto_slice_of_cpp_vec = rust::Slice(cpp_vec);
warning: cxx-test-suite@0.0.0:       |                                                     ^            
warning: cxx-test-suite@0.0.0: /usr/local/google/home/jrreinhart/code/cxx/target/debug/build/cxx-test-suite-409adf2eee31b672/out/cxxbridge/include/rust/cxx.h:565:1: error: forming pointer to reference type 'int&'                           
warning: cxx-test-suite@0.0.0:   565 | Slice<T>::Slice(T *s, std::size_t count) noexcept {                                                                                                                                                     
warning: cxx-test-suite@0.0.0:       | ^~~~~~~~                                                                        
warning: cxx-test-suite@0.0.0: /usr/local/google/home/jrreinhart/code/cxx/target/debug/build/cxx-test-suite-409adf2eee31b672/out/cxxbridge/include/rust/cxx.h:575:4: error: forming pointer to reference type 'int&'                           
warning: cxx-test-suite@0.0.0:   575 | T *Slice<T>::data() const noexcept {                                                                                                                                                                    
warning: cxx-test-suite@0.0.0:       |    ^~~~~~~~                                                                                                                                                                                             
warning: cxx-test-suite@0.0.0: tests.cc: In function 'const char* tests::cxx_run_test()':                                                                                                                                                      
warning: cxx-test-suite@0.0.0: tests.cc:910:14: error: static assertion failed                                         
warning: cxx-test-suite@0.0.0:   910 |         std::is_same_v<decltype(auto_slice_of_cpp_vec), rust::Slice<int>>);     
warning: cxx-test-suite@0.0.0:       |         ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                                                                                                                                                              
warning: cxx-test-suite@0.0.0: tests.cc:911:34: error: 'class rust::cxxbridge1::Slice<int&>' has no member named 'data'; did you mean 'at'?                                                                                                    
warning: cxx-test-suite@0.0.0:   911 |     ASSERT(auto_slice_of_cpp_vec.data() == cpp_vec.data());                                                                                                                                             
warning: cxx-test-suite@0.0.0:       |                                  ^~~~                                                                                                                                                                   
warning: cxx-test-suite@0.0.0: tests.cc:765:11: note: in definition of macro 'ASSERT'                                  
warning: cxx-test-suite@0.0.0:   765 |     if (!(x)) {                                                                \                                                                                                                        
warning: cxx-test-suite@0.0.0:       |           ^                                                                                                                                                                                             
warning: cxx-test-suite@0.0.0: /usr/local/google/home/jrreinhart/code/cxx/target/debug/build/cxx-test-suite-409adf2eee31b672/out/cxxbridge/include/rust/cxx.h: In instantiation of 'class rust::cxxbridge1::Slice<const int&>':                
warning: cxx-test-suite@0.0.0: tests.cc:917:53:   required from here                                                                                                                                                                           
warning: cxx-test-suite@0.0.0:   917 |     auto auto_slice_of_cpp_vec = rust::Slice(cpp_vec);
warning: cxx-test-suite@0.0.0:       |                                                     ^ 
warning: cxx-test-suite@0.0.0: /usr/local/google/home/jrreinhart/code/cxx/target/debug/build/cxx-test-suite-409adf2eee31b672/out/cxxbridge/include/rust/cxx.h:565:1: error: forming pointer to reference type 'const int&'                     
warning: cxx-test-suite@0.0.0:   565 | Slice<T>::Slice(T *s, std::size_t count) noexcept {                                                                                                                                                     
warning: cxx-test-suite@0.0.0:       | ^~~~~~~~                                                                        
warning: cxx-test-suite@0.0.0: /usr/local/google/home/jrreinhart/code/cxx/target/debug/build/cxx-test-suite-409adf2eee31b672/out/cxxbridge/include/rust/cxx.h:575:4: error: forming pointer to reference type 'const int&'                     
warning: cxx-test-suite@0.0.0:   575 | T *Slice<T>::data() const noexcept {                                                                                                                                                                    
warning: cxx-test-suite@0.0.0:       |    ^~~~~~~~                                                                     
warning: cxx-test-suite@0.0.0: tests.cc:918:24: error: static assertion failed                                         
warning: cxx-test-suite@0.0.0:   918 |     static_assert(std::is_same_v<decltype(auto_slice_of_cpp_vec),
warning: cxx-test-suite@0.0.0:       |                   ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~               
warning: cxx-test-suite@0.0.0:   919 |                                  rust::Slice<const int>>);                      
warning: cxx-test-suite@0.0.0:       |                                  ~~~~~~~~~~~~~~~~~~~~~~~                        
warning: cxx-test-suite@0.0.0: tests.cc:920:34: error: 'class rust::cxxbridge1::Slice<const int&>' has no member named 'data'; did you mean 'at'?                                                                                              
warning: cxx-test-suite@0.0.0:   920 |     ASSERT(auto_slice_of_cpp_vec.data() == cpp_vec.data());                                                                                                                                             
warning: cxx-test-suite@0.0.0:       |                                  ^~~~                                           
warning: cxx-test-suite@0.0.0: tests.cc:765:11: note: in definition of macro 'ASSERT'        
warning: cxx-test-suite@0.0.0:   765 |     if (!(x)) {                                                                \                                                                                                                        
warning: cxx-test-suite@0.0.0:       |           ^                                                                                                                                                                                             
```

anforowicz: > Can you explain why `std::remove_reference_t` is necessary?
> 
> I tried without that, and the tests did try to create `Slice<int&>`. But I don't understand where the reference is coming from, since I would expect `*std::declval<C>().data()` to just be a `T` (since the signature is `T* data()`).
 
IIUC `*std::declval<C>().data()` is `T&`.  Not sure if a simplified example here helps: https://godbolt.org/z/njoYqz8ez

dtolnay (approved):
Thank you!

PR #1454, opened 2025-03-03
`impl<T> Seek for UniquePtr<T> where ... Pin<&a mut T> : Seek`.

anforowicz:
This commit implements forwarding of `Seek` trait implementation from `UniquePtr<T>` to the pointee type.  This is quite similar to how `Box<T>` also forwards - see
https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-Seek-for-Box%3CS%3E

This commit has quite similar, orphan-rule-related motivation as the earlier https://github.com/dtolnay/cxx/pull/1368 which covered the `Read` trait.  A more specific motivating example is https://crbug.com/400455848 where `png` crate changed the old `Read` constraint to `Read + Seek + ...`.

### DIFF
--- a/src/unique_ptr.rs
+++ b/src/unique_ptr.rs
@@ -14,7 +14,7 @@ use core::mem::{self, MaybeUninit};
 use core::ops::{Deref, DerefMut};
 use core::pin::Pin;
 #[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #1458, opened 2025-03-04
Switch to standard feature-test macros

dtolnay:
https://en.cppreference.com/w/cpp/feature_test

This matches what was done in https://github.com/dtolnay/cxx/pull/1449 for deduction guides.


dtolnay:
For this:

https://github.com/dtolnay/cxx/blob/03b025dd0380a4845a3ab26497691d47e322e2db/include/cxx.h#L232-L236

I did not find the right feature test macro to use for `std::contiguous_iterator_tag`.

PR #1459, opened 2025-03-05
Add support for `rust::Option<Box<T>>`

djmitche:
While general support for `Option<T>` is difficult, `Option<Box<T>>` has a well-known layout due to the [NPO](https://doc.rust-lang.org/std/option/#representation), so we can replicate that layout in C++.

---

This is a very early draft!

### DIFF
--- a/gen/src/builtin.rs
+++ b/gen/src/builtin.rs
@@ -10,6 +10,7 @@ pub(crate) struct Builtins<'a> {
     pub rust_slice: bool,
     pub rust_box: bool,
     pub rust_vec: bool,
+    pub rust_option: bool,
     pub rust_fn: bool,
     pub rust_isize: bool,
     pub opaque: bool,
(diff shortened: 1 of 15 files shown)


djmitche:
I'd like a little feedback here. I've looked for the places where `vec` was referenced and tried to do similar things, although it turns out the implementation is closer to `rust::Box`. My understanding of what I'm doing is based on `git grep` and passing tests, and is not at all comprehensive.

 * Is this on the right track? Asking another way, do you look at this and thing "what is this nonsense" or "ok, it's a start"?
 * What issues, soundness or otherwise, have I not addressed?
   * One that comes to mind is ZSTs, which have non-zero pointer values that cannot be dereferenced. But, I think those are not allowed in `rust::Box` anyway?
 * On what should I model the methods for `rust::Option` -- `std::Option` or `std::optional`? or neither?
 * What else needs done? I know I need to write a .html page, and there are a few TODOs in the PR that I will get to.
   * I do need to add tests for Rust returning `Option<Box<T>>`.

djmitche:
Well, in the absence of any negative feedback, I guess I'll put this up for review :)

djmitche:
Withdrawn for lack of interest

Issue #26, opened 2020-01-26
Consider emitting cargo:rerun-if-changed line for argument of cxx::Build::bridge


dtolnay:
https://docs.rs/cxx/0.1.2/cxx/struct.Build.html#method.bridge could automatically do:

```rust
println!("cargo:rerun-if-changed={}", rust_source_file);
```

According to https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script:

> `rerun-if-changed=PATH` is a path to a file or directory which indicates that the build script should be re-run if it changes (detected by a more-recent last-modified timestamp on the file). Normally build scripts are re-run if any file inside the crate root changes, but this can be used to scope changes to just a small set of files.

dtolnay:
I am leaning against doing this because the pit of success is bigger if we don't. If we don't emit rerun-if-changed, the downsides are:

- build.rs might be rerun by Cargo more often than necessary, whenever anything under the crate's root directory changes;
- the user must add their own rerun-if-changed line in their build.rs to avoid those reruns.

If we do emit rerun-if-changed:

- and the user's build script does something *other than* cxx build, then those other parts no longer rerun automatically when anything under the crate's root directory changes, and it's hard to discover why.

PR #1462, opened 2025-03-06
Auto-track all bridge files

### LINKED ISSUE #1461: Automatically generate `cargo:rerun-if-changed`
I think by now the `cargo:rerun-if-changed=<path>` is ubiqutous in all cxx-using cases, so I think it would make a lot of sense to automatically generate these as part of the bridge building.  This is especially important because most of the time the path string is hardcoded in the user's code, so may be easily mismatch and cause surprising behavior.

nyurik:
Adds `cargo:rerun-if-changed=<path>` for all files used by the bridge.

Note that when I went through all the `rerun-if-changed` cases, some of them did **not** include the .rs file - which may mean that even this repo is missing the needed tracking and this would fix those.

Fixes #1461

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,6 @@ fn main() {
         .std("c++11")
         .compile("cxxbridge-demo");
 
-    println!("cargo:rerun-if-changed=src/main.rs");
     println!("cargo:rerun-if-changed=src/demo.cc");
     println!("cargo:rerun-if-changed=include/demo.h");
 }
--- a/book/src/build/cargo.md
+++ b/book/src/build/cargo.md
@@ -41,7 +41,6 @@ fn main() {
         .std("c++11")
         .compile("cxxbridge-demo");
 
-    println!("cargo:rerun-if-changed=src/main.rs");
     println!("cargo:rerun-if-changed=src/demo.cc");
     println!("cargo:rerun-if-changed=include/demo.h");
 }
--- a/book/src/tutorial.md
+++ b/book/src/tutorial.md
@@ -205,7 +205,6 @@ fn main() {
         .file("src/blobstore.cc")
         .compile("cxx-demo");
 
-    println!("cargo:rerun-if-changed=src/main.rs");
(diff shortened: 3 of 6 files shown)


dtolnay (commented):
See previous consideration of this in https://github.com/dtolnay/cxx/issues/26. I think it's better not to do this.

nyurik:
@dtolnay thx, i didn't see that PR.   Do you think cxx could adapt the approach to other similar crates like `static_files` [change_detection](https://docs.rs/static-files/latest/static_files/struct.NpmBuild.html#method.change_detection)?  It has a builder that allows to enable change detction print statements, and afterwards it finalizes it.

The issue with cxx is that it produces a `cc::Build` directly - making it somewhat inconvinient when additional configurations are needed.  Perhaps we can introduce a new `Bridge` or `CxxBuild` struct?

```rust
    cxx_build::new()  // empty CxxBuild
        .bridge("src/map_renderer.rs") // add a single bridge file
        .change_detection() // print change detections
        .bridge()  // consume CxxBuild and create cc::Build
        .includes(&include_dirs)  // Just as before - operations on the cc::Build
        .file("src/renderer/bridge.cpp")
        .flag_if_supported("-std=c++20")
        .compile("maplibre_rust_map_renderer_bindings");
```

nyurik:
Thx, I created https://github.com/dtolnay/cxx/pull/1463 to do that

dtolnay:
Beginning in cc version 1.0.74 the cc crate has started printing some rerun-if-env-changed lines by default, so the downsides in https://github.com/dtolnay/cxx/issues/26 are already being paid. (From the description of the relevant cc PR it sounds like the downside was not understood when making that behavior change.) In that case this PR is fine after all and will not make things worse.

phil-opp:
> For example do not do this for files located in OUT_DIR

Even better, thanks for the quick reply! I opened a PR at https://github.com/dtolnay/cxx/pull/1668.

PR #1463, opened 2025-03-06
Introduce `CFG.change_detection` bool flag

### LINKED ISSUE #1461: Automatically generate `cargo:rerun-if-changed`

nyurik:
Adds `cargo:rerun-if-changed=<path>` for all files used by the bridge if the `CFG.change_detection` is `true`.

Per https://github.com/dtolnay/cxx/pull/1462#issuecomment-2704977488 
Fixes #1461

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,12 +242,12 @@ cxx-build = "1.0"
 // build.rs
 
 fn main() {
+    cxx_build::CFG.change_detection = true;
     cxx_build::bridge("src/main.rs")  // returns a cc::Build
         .file("src/demo.cc")
         .std("c++11")
         .compile("cxxbridge-demo");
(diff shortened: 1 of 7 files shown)


nyurik:
@dtolnay does this implement what you meant in your comment?

dtolnay (commented):
Thanks &mdash; this implementation is what I meant but I ended up merging #1462 after all.

PR #1464, opened 2025-03-10
macro: ensure that cfg attrs are set for type alias verify generation

ahayzen-kdab:
```rust
mod ffi {
  unsafe extern "C++" {
    #[cfg(enabled)]
    type A = crate::A;
  }
}
```

When using the bridge above we need to ensure that all generated code also has the cfg attribute set. Before this change the following error would occur `cannot find type A in this scope`. This appears to be due to the generated verify block below that was not copying the attributes from the original type.

```rust
const _: fn() = ::cxx::private::verify_external_type::<A, ...>;
````

After this change the attributes are copied and therefore when the type is not compiled the verify block is also not compiled.

### DIFF
--- a/macro/src/expand.rs
+++ b/macro/src/expand.rs
@@ -1264,6 +1264,7 @@ fn expand_type_alias(alias: &TypeAlias) -> TokenStream {
 }
 
 fn expand_type_alias_verify(alias: &TypeAlias, types: &Types) -> TokenStream {
+    let attrs = &alias.attrs;
     let ident = &alias.name.rust;
     let type_id = type_id(&alias.name);
     let begin_span = alias.type_token.span;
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Can you confirm that you have tested that this works? I would not expect cfg attributes to be present in `attrs` so this PR as currently written should have no effect on what you are describing.

ahayzen-kdab:
I could see the cfg attribute appearing via `cargo expand`, but i would personally prefer this to be in a test to be sure as well.

@dtolnay do you have any recommendations as to where i could write a test for this ? As i had a quick look but it wasn't obvious if i should put this in `tests/` or in an inline `#[test]` etc.

If you could point me to an example of a test taking a Rust bridge as input and asserting the generated Rust output that would be useful.

dtolnay:
Here is where I got mixed up: I forgot that cfg attributes end up going into _both_ `cfg` and `attrs` (unlike some other attributes, which are partitioned into only the Doc, or only the Vec\<Derive\>, or only the OtherAttrs). The PR makes perfect sense now. Thank you!

https://github.com/dtolnay/cxx/blob/82765ef1cdccbe6af51225c927dada8e1508dade/syntax/mod.rs#L167-L174

https://github.com/dtolnay/cxx/blob/82765ef1cdccbe6af51225c927dada8e1508dade/syntax/attrs.rs#L132-L139

dtolnay (approved):

dtolnay:
Published in 1.0.144.

PR #1466, opened 2025-03-13
Updated bzlmod to support Bazel 8

UebelAndre:
This also pulls in some modernization where `run_binary` can now support `execpath` expansion and through the use of [bazel-features](https://github.com/bazel-contrib/bazel_features) we're able to reduce the size of `MODULE.bazel.lock` files by flagging `cxx` dependencies as reproducible.


dtolnay (approved):
Thank you!

Issue #1095, opened 2022-09-14
Publish to Bazel Central Registry


dtolnay:
With a suitable MODULE.bazel, we should be able to make a `rust_cxx_bridge` rule easily available for Bzlmod users.

https://github.com/bazelbuild/bazel-central-registry
https://bazel.build/build/bzlmod

For now this is blocked on Bzlmod support in https://github.com/bazelbuild/rules_rust/issues/1493.

tel:
Given #1294 has merged, is it possible that this will get published to the central registry?

I've been trying to get this working using `git_override` and struggling a bit, though I'm pretty new to Bazel. I'm curious if this is an intended way to use cxx in a Bazel project.

louwers:
@tel I was able to use cxx with `git_override` like this: https://github.com/maplibre/maplibre-native/blob/527e70b6fa7f2ff1d3c7546f1773c5404b8a3733/MODULE.bazel#L65-L70

UebelAndre:
I made https://github.com/dtolnay/cxx/pull/1466 so that the repo could be directly published going forward.

PR #1467, opened 2025-03-13
Added support for 'publish-to-bcr' tool

UebelAndre:
This change should add all the necessary components to support [bazel-contrib/publish-to-bcr](https://github.com/bazel-contrib/publish-to-bcr). All @dtolnay would need to do is follow the directions for that tool to setup automatic Bazel Central Registry deploys. Which I think is forking https://github.com/bazelbuild/bazel-central-registry and adding the app to their github account.

closes https://github.com/dtolnay/cxx/issues/1095

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,0 +1,15 @@
+matrix:
+  platform:
+    - macos_arm64
+    - ubuntu2404
+    - windows
+  bazel: [7.x, 8.x]
+tasks:
+  verify_targets:
+    name: Verify build targets
+    platform: ${{ platform }}
+    bazel: ${{ bazel }}
+    build_targets:
+      - '//...'
+    test_targets:
+      - '//...'
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,6 +1,6 @@
 module(
     name = "cxx.rs",
-    version = "1.0.145",
+    version = "0.0.0",
     bazel_compatibility = [">=7.2.1"],
     compatibility_level = 1,
 )
--- a/.bcr/README.md
+++ b/.bcr/README.md
@@ -1,0 +1,9 @@
+# Bazel Central Registry
+
+When the ruleset is released, we want it to be published to the
+Bazel Central Registry automatically:
+<https://registry.bazel.build>
+
+This folder contains configuration files to automate the publish step.
+See <https://github.com/bazel-contrib/publish-to-bcr/blob/main/templates/README.md>
+for authoritative documentation about these files.
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,0 +1,5 @@
+fixedReleaser:
+  login: dtolnay
+  email: 1940490+dtolnay@users.noreply.github.com
+moduleRoots:
+  - "."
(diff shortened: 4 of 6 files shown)


.bcr/presubmit.yml
```diff
@@ -0,0 +1,15 @@
+matrix:
+  platform:
+    - macos_arm64
+    - ubuntu2404
+    - windows
```
dtolnay, line 5: Could you look into what has been going wrong in the Windows case?
https://github.com/dtolnay/cxx/actions/runs/13841971991/job/38731434026

> `ERROR: C:/users/runneradmin/_bazel_runneradmin/dzh43mlk/external/rules_rust+/util/process_wrapper/BUILD.bazel:31:36: Compiling Rust (without process_wrapper) bin process_wrapper (6 files) [for tool] failed: (Exit 1): bootstrap_process_wrapper.bat failed: error executing Rustc command (from target @@rules_rust+//util/process_wrapper:process_wrapper)`

> `LINK : fatal error LNK1181: cannot open input file 'bazel-out\x64_windows-opt-exec-ST-d57f47055a04\bin\external\rules_rust++rust+rw-1704990954_tools\rust_toolchain\lib\rustlib\x86_64-pc-windows-msvc\lib\librustc_std_workspace_alloc-2cc5ffc24865e22c.rlib'`

UebelAndre: I opened https://github.com/dtolnay/cxx/pull/1468 for this

UebelAndre: I would also be happy to drop windows for now until the underlying issues are resolved. This doesn't prevent anyone from pulling the module on windows, just means windows won't be tested for releases to BCR.
```suggestion
```

MODULE.bazel
```diff
@@ -1,6 +1,6 @@
 module(
     name = "cxx.rs",
-    version = "1.0.145",
+    version = "0.0.0",
```
dtolnay, line 3: What is the significance of this version number?

- Does it become the directory name in bazel-central-repository? https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/cxx.rs, or is that determined by the git tag when I publish a release?

- Is it what users need to write in `bazel_dep` in their downstream MODULE.bazel?

UebelAndre: > Does it become the directory name in bazel-central-registry? https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/cxx.rs, or is that determined by the git tag when I publish a release?

I believe the way https://github.com/bazel-contrib/publish-to-bcr works is based on github releases. When you cut a release, this version will be updated to match no matter what it is. Thus, to try and avoid the toil if needing to update this whenever you do a release, I thought it would be useful to just set it to `0.0.0` (similar to [rules_python](https://github.com/bazelbuild/rules_python/blob/389431bba6f9a4b46b6cf15dd9cd24a1f52f6e16/MODULE.bazel#L3))

> * Is it what users need to write in `bazel_dep` in their downstream MODULE.bazel?

Yes, but again, this value is replaced upon cutting a release (or at least that's my understanding).

dtolnay: Sounds good!

UebelAndre:
> Thank you!
> 
> I have configured https://github.com/apps/publish-to-bcr with access to this repo + https://github.com/dtolnay/bazel-central-registry, and I will try publishing a release right after merging this.

I noticed the new release and no pull-request in BCR so I'm asking on Slack. The tool is still kinda a mystery to me.

PR #1470, opened 2025-03-14
Align Bazel preferred linkages with Buck

dtolnay:
Everywhere that Buck is using `preferred_linkage = "static"`, Bazel is going to need `linkstatic = True` for the same reason. This matches Cargo as well, which statically links C dependencies into the same crate whose build.rs builds them.

### DIFF
--- a/tools/bazel/rust_cxx_bridge.bzl
+++ b/tools/bazel/rust_cxx_bridge.bzl
@@ -3,7 +3,7 @@
 load("@bazel_skylib//rules:run_binary.bzl", "run_binary")
 load("@rules_cc//cc:defs.bzl", "cc_library")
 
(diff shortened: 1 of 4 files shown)


tools/bazel/rust_cxx_bridge.bzl
```diff
@@ -46,6 +46,7 @@ def rust_cxx_bridge(name, src, deps = [], **kwargs):
         name = name,
         srcs = [src + ".cc"],
         deps = deps + [":%s/include" % name],
+        linkstatic = linkstatic,
```
dtolnay, line 49: https://github.com/dtolnay/cxx/blob/dcc80b53b0d2863b18d8cb900d376289fff577b3/tools/buck/rust_cxx_bridge.bzl#L28-L33

PR #1471, opened 2025-03-14
chore: cleanup a few format args

nyurik:
Make some code a bit more readable with inlined format args (some code is already using them, so might as well do some more now that the 2021 edition is everywhere)

Some readability examples:

```rust
format!("{}({{}})", ident);   /* --> */  format!("{ident}({{}})");
```


dtolnay (commented):
Thanks for the PR!

I am not that fond of this style. I think a few of these are better this way, and a few are better the other way, and the rest there is no readability improvement in either direction. Overall I think the existing code is all right as is.

PR #1480, opened 2025-03-15
Include Bazel on Windows in vendored targets verification

dtolnay:


dtolnay (commented):
Wow, this is extremely slow. Linux: 5m49s. macOS: 6m01s. Windows: 12m05s.

dtolnay:
It looks like most of the time is spent compiling https://github.com/GitoxideLabs/gitoxide, which is a dependency of `crates-index`, which is needed by `cargo-bazel`.

PR #1486, opened 2025-03-24
Add an explicit conversion operator from `rust::Str` to `std::string_view`.

anforowicz:
PTAL?

I am aware of https://github.com/dtolnay/cxx/pull/80 (adding a way to construct `rust::Str` from `std::string_view`, using a custom C++17-detection mechanism) and https://github.com/dtolnay/cxx/pull/410 (adding a way to construct `rust::String` from `std::string_view`, also using a custom C++17 detection mechanism).  But maybe we can still proceed with this PR given that:

* This PR is relatively simple - just adds one member (a conversion operator) to `rust::Str`.  I _think_ that adding such an **explicit** operator has close to zero risk of breaking existing users (unlike the other PRs where new constructor overloads may risk ambiguity or other resolution problems)
* Since the other PRs were raised, we were able to proceed with other C++-version-dependent APIs:
    - https://github.com/dtolnay/cxx/pull/1432 added a (C++-20-specific) way to convert `rust::Slice` into Chromium's `base::span` (which needed `rust::Slice` to expose a different `iterator_category` in C++20 than in earlier versions).
    - https://github.com/dtolnay/cxx/pull/1437 added CI coverage for different C++ versions.

FWIW I've also noticed https://github.com/dtolnay/cxx/commit/5ae3a93b425abb2205e718302a033c688260f639 where `#if __cplusplus >= 202002L` was replaced with `#ifdef __cpp_char8_t`.  IIUC this is using https://en.cppreference.com/w/cpp/utility/feature_test which are available in C++20 and later.  So I think we can't do that in this PR.

### DIFF
--- a/book/src/binding/str.md
+++ b/book/src/binding/str.md
@@ -25,6 +25,9 @@ public:
   Str &operator=(const Str &) & noexcept;
 
   explicit operator std::string() const;
+#if __cplusplus >= 201703L
+  explicit operator std::string_view() const;
+#endif
 
   // Note: no null terminator.
   const char *data() const noexcept;
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -9,9 +9,6 @@
 #include <iosfwd>
 #include <iterator>
 #include <new>
-#if __cplusplus >= 202002L
-#include <ranges>
-#endif
 #include <stdexcept>
 #include <string>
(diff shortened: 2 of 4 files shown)


dtolnay (approved):
Thanks!

PR #1496, opened 2025-04-06
Fill in GitHub user ID for BCR publisher

dtolnay:
https://github.com/bazelbuild/bazel-central-registry/blob/d561ba93f5ef84a15e5fbe086052f68e56fcea1b/metadata.schema.json

If the dtolnay user is deleted from GitHub, this PR prevents someone who takes over that username from publishing releases of this crate to Bazel Central Registry.

### DIFF
--- a/.bcr/metadata.template.json
+++ b/.bcr/metadata.template.json
@@ -3,6 +3,7 @@
   "maintainers": [
     {
       "github": "dtolnay",
+      "github_user_id": 1940490,
       "email": "dtolnay@gmail.com",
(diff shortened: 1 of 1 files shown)


.bcr/metadata.template.json
```diff
@@ -3,6 +3,7 @@
   "maintainers": [
     {
       "github": "dtolnay",
+      "github_user_id": 1940490,
```
dtolnay, line 6: https://api.github.com/users/dtolnay

Issue #1497, opened 2025-04-08
Recent zero-initialization triggers `-Wreturn-type-c-linkage` when building generated code with `clang++`


anforowicz:
During a recent attempt to roll `cxx` to a new version in Chromium, we discovered that the code generated by `cxx` results in a Clang warning that didn't happen before.  I am opening an issue to capture what I learned so far, but I admit that I didn't really dig deeper into the issue...  but I _assume_ that the problem pointed out by the warning is real?

REPRO:

* `export CXXFLAGS="-Werror -Wall -std=c++14"`
    - C++14 because `CXX_DEFAULT_VALUE` expands to "nothing" in earlier C++ versions - see [here](https://github.com/dtolnay/cxx/blob/dfecf55cb2dfb825a6c2541b6f43d06040c9c41a/gen/src/write.rs#L51-L55)
    - `-Werror` and `-Wall` to error out if `-Wreturn-type-c-linkage` is reported
* `export CXX=clang++`
    - The problem doesn't repro with the default `c++` compiler
    - The problem originally repro-ed with Chromium-provided compiler (nightly version of Clang), but it also repros with what I have installed on my system:
 
        ```
        $ clang++ --version
        Debian clang version 19.1.7 (1)
        Target: x86_64-pc-linux-gnu
        Thread model: posix
        InstalledDir: /usr/lib/llvm-19/bin
        ```
* `git checkout ead032256713c6aa0de3b410417357b3e44ce5ab`
    - ead032256713c6aa0de3b410417357b3e44ce5ab is what the problem has bisected to.  The problem doesn't repro at an earlier commit.
* `cargo test`

```
$ cargo test
   Compiling cxx-test-suite v0.0.0 (/usr/local/google/home/lukasza/src/github/cxx/tests/ffi)
warning: cxx-test-suite@0.0.0: /usr/local/google/home/lukasza/src/github/cxx/target/debug/build/cxx-test-suite-6288a940afe915f9/out/cxxbridge/sources/tests/ffi/lib.rs.cc:1471:17: error: 'tests$cxxbridge1$c_return_shared' has C-linkage specified, but returns user-defined type '::tests::Shared' which is incompatible with C [-Werror,-Wreturn-type-c-linkage]
warning: cxx-test-suite@0.0.0:  1471 | ::tests::Shared tests$cxxbridge1$c_return_shared() noexcept {
warning: cxx-test-suite@0.0.0:       |                 ^
[...]

PR #1499, opened 2025-04-08
Add `-Wall` and `-Werror` to `CXXFLAGS` used by GitHub CI.

anforowicz:
Let me start this as a **draft** PR to double-check that the CI passes with the new flags.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,15 +35,15 @@ jobs:
           - name: C++14
             rust: nightly
             os: ubuntu
-            flags: -std=c++14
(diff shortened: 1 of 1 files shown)


anforowicz:
Ok, this seems to be ready for a review - can you PTAL?

Motivation for this is to try to catch issues like https://github.com/dtolnay/cxx/issues/1497.  This PR on its own may be insufficient (because it doesn't install/force usage of Clang), but it seems like a step in the right direction.  I'll try to also cover Clang in a follow-up.

dtolnay (approved):

PR #1527, opened 2025-06-24
Remove `#[derive(Copy, Clone)]` from `NamedImpl`

anforowicz:
This PR is a refactoring (i.e. it doesn't _directly_ fix any bugs or add any features) that simplifies the code a little bit:

* First commit: fix test expectations to match error messages from the latest nightly version of `rustc`
* Second commit: delete most code from `OrderedMap` and instead reuse `indexmap` crate
    - The main motivation here is that `indexmap::IndexMap<K, V>` doesn't have `K: Copy` constraint unlike the old implementation of `OrderedMap`.
    - A side effect of the main change is that some internal APIs become no longer needed (e.g. `fn iter`)
* Third commit: remove `#[derive(Copy, Clone)]` from `NamedImpl`

This PR is secretly part of a bigger work item (trying to add support for `Vec<[u8; 16]>`), but that bigger work item is... well... bigger (and not well fleshed out at this point), so I hope to land these 3 (hopefully non-controversial) commits for now.


dtolnay (approved):
Thanks!

Issue #188, opened 2020-05-05
Add test harness for testing false C++ static_asserts


dtolnay:
In #187 we wanted to have testing that some `static_assert` correctly evaluates to false when intended and that a particular error message is shown.

```cpp
enum class E : uint32_t {
    A = 1,
};
```

```rust
mod ffi {
    enum E {
        A = 2,
    }

    extern "C" {
        type E;  // generated C++ code should fail a static_assert
    }
}
```

PR #1536, opened 2025-07-17
Make `is_complete` static assertion unconditional for `unique_ptr<T>` bindings. 

anforowicz:
PTAL?

FWIW I tried to leave a justification of small design/coding decisions here and there in code comments + in the description of the last commit.

Similarly to my last PR, this one is also secretly part of a bigger work item (*).  I hope that when the bigger work item is done, then `enum UniquePtr`, `trait ToTypename`, and `trait ToMangled` can be removed altogether from `gen/src/write.rs`.  But even if this doesn't materialize, the changes from **this** PR under review seem desirable on their own (adding test coverage for C++-side errors + simplifying `fn write_unique_ptr_common` and removing _some_ uses/deconstructions of `UniquePtr`).

WDYT?

(*) "bigger work item" has evolved from trying to add support for `Vec<[u8; N]>` to adding support for `Vec<T>` where `T` is not necessarily a simple identifier (e.g. for `Vec<Vec<T>>`).  I am still working at this point on the last few remaining places where `write.rs` or `expand.rs` only work with identifier-only/named types (one of those places is `conditional_delete`, another is `is_maybe_trivial`, and `fn write_unique_ptr_common` is another one).


anforowicz:
Force push above rebases on top of the upstream origin/master - this should make this PR mergeable again.

anforowicz:
It seems that this PR will fix https://github.com/dtolnay/cxx/issues/188

dtolnay (approved):
Thank you!

PR #1539, opened 2025-07-25
Reusing Rust type bindings / support for type aliases in the Rust section

anforowicz:
PTAL?

This PR fixes https://crbug.com/433254489 (/cc @shabbyx as FYI - IIUC they've patched this PR onto their local machine and verified that it does indeed unblock them).

The first commit fixes some existing code duplication and prevents more code duplication around how `fn syntax::parse_items` consumes a `syntax::Module`.  I think this commit is okay and moves things in a reasonable direction, but I acknowledge that adding more partial consumption/destructuring/taking is a bit icky.  More discussion about this aspect of the changes can be found in a comment from a semi-internal review at https://github.com/anforowicz/cxx/pull/2#discussion_r2229465883 and https://github.com/anforowicz/cxx/pull/2#discussion_r2231757082.

I also note that some existing variants of `syntax::Api` `enum` have Rust and C++ versions - e.g. `RustFunction` and `CxxFunction` as well as `RustType` and `CxxType`.  We could also do that for the `TypeAlias` variant which this PR doesn't touch just yet.  Please provide feedback on whether I should also submit an additional commit into this or a separate PR - see https://github.com/anforowicz/cxx/tree/api-rust-type-alias-separate-variant

/cc @zetafunction who has kindly provided initial, semi-internal feedback at https://github.com/anforowicz/cxx/pull/2

### DIFF
--- a/book/src/extern-rust.md
+++ b/book/src/extern-rust.md
@@ -194,3 +194,46 @@ mod ffi {
 
 Bounds on a lifetime (like `<'a, 'b: 'a>`) are not currently supported. Nor are
 type parameters or where-clauses.
+
+## Reusing existing binding types
+
+Extern Rust types support a syntax for declaring that a C++ binding of the
+given Rust type already exists outside of the current bridge module. This
+avoids generating a separate binding which would clash with the first one
+(e.g. one possible error is: "conflicting implementations of trait `RustType`").
+
+The snippet below illustrates how to unify references to a Rust type
+across bridges:
+
+```rs
+// file1.rs
+#[cxx::bridge(namespace = "example")]
+pub mod ffi {
+    extern "Rust" {
+        type Demo;
+
+        fn create_demo() -> Box<Demo>;
+    }
+}
+
+pub struct Demo;
+
+pub fn create_demo() -> Box<Demo> { todo!() }
+```
+
+```rs
+// file2.rs
+#[cxx::bridge(namespace = "example")]
+pub mod ffi {
+    extern "C++" {
+        include!("file1.rs.h");
+    }
+    extern "Rust" {
+        type Demo = crate::file1::Demo;
+
+        fn take_ref_demo(demo: &Demo);
+    }
+}
+
+pub fn take_ref_demo(demo: &crate::file1::Demo) { todo!() }
+```
--- a/macro/src/expand.rs
+++ b/macro/src/expand.rs
@@ -1339,26 +1339,37 @@ fn expand_type_alias(alias: &TypeAlias) -> TokenStream {
 fn expand_type_alias_verify(alias: &TypeAlias, types: &Types) -> TokenStream {
     let attrs = &alias.attrs;
     let ident = &alias.name.rust;
-    let type_id = type_id(&alias.name);
     let begin_span = alias.type_token.span;
     let end_span = alias.semi_token.span;
-    let begin = quote_spanned!(begin_span=> ::cxx::private::verify_extern_type::<);
     let end = quote_spanned!(end_span=> >);
 
-    let mut verify = quote! {
-        #attrs
-        const _: fn() = #begin #ident, #type_id #end;
-    };
+    match alias.lang {
+        Lang::Cxx | Lang::CxxUnwind => {
+            let type_id = type_id(&alias.name);
+            let begin = quote_spanned!(begin_span=> ::cxx::private::verify_extern_type::<);
+            let mut verify = quote! {
+                #attrs
+                const _: fn() = #begin #ident, #type_id #end;
+            };
 
-    if types.required_trivial.contains_key(&alias.name.rust) {
-        let begin = quote_spanned!(begin_span=> ::cxx::private::verify_extern_kind::<);
-        verify.extend(quote! {
-            #attrs
-            const _: fn() = #begin #ident, ::cxx::kind::Trivial #end;
-        });
+            if types.required_trivial.contains_key(&alias.name.rust) {
+                let begin = quote_spanned!(begin_span=> ::cxx::private::verify_extern_kind::<);
+                verify.extend(quote! {
+                    #attrs
+                    const _: fn() = #begin #ident, ::cxx::kind::Trivial #end;
+                });
+            }
+
+            verify
+        }
+        Lang::Rust => {
+            let begin = quote_spanned!(begin_span=> ::cxx::private::verify_rust_type::<);
+            quote! {
+                #attrs
+                const _: fn() = #begin #ident #end;
+            }
+        }
     }
-
-    verify
 }
 
 fn type_id(name: &Pair) -> TokenStream {
--- a/macro/src/expand.rs
+++ b/macro/src/expand.rs
@@ -34,10 +34,7 @@ pub(crate) fn bridge(mut ffi: Module) -> Result<TokenStream> {
         },
     );
 
-    let content = mem::take(&mut ffi.content);
-    let trusted = ffi.unsafety.is_some();
-    let namespace = &ffi.namespace;
-    let ref mut apis = syntax::parse_items(errors, content, trusted, namespace);
+    let ref mut apis = syntax::parse_items(errors, &mut ffi);
     let ref types = Types::collect(errors, apis);
     errors.propagate()?;
 
--- a/book/package-lock.json
+++ b/book/package-lock.json
@@ -59,13 +59,12 @@
       }
     },
     "node_modules/@eslint/config-array": {
-      "version": "0.19.1",
-      "resolved": "https://registry.npmjs.org/@eslint/config-array/-/config-array-0.19.1.tgz",
-      "integrity": "sha512-fo6Mtm5mWyKjA/Chy1BYTdn5mGJoDNjC7C64ug20ADsRDGrA85bN3uK3MaKbeRkRuuIEAR5N33Jr1pbm411/PA==",
+      "version": "0.21.0",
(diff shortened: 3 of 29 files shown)


book/src/extern-rust.md
```diff
@@ -163,3 +163,46 @@ mod ffi {
 
 Bounds on a lifetime (like `<'a, 'b: 'a>`) are not currently supported. Nor are
 type parameters or where-clauses.
+
+## Reusing existing binding types
```
dtolnay, line 167: Reusing an existing defined pair of C++ and Rust type is already supported. I don't understand the distinction between the extern C++ type aliases which are already supported, and the extern Rust type aliases in this PR. What is the situation where it would be correct to write `extern "Rust" { type T = path::to::U; }` and not `extern "C++" { type T = path::to::U; }`?

macro/src/expand.rs
```diff
@@ -1273,26 +1270,37 @@ fn expand_type_alias(alias: &TypeAlias) -> TokenStream {
...
+                verify.extend(quote! {
+                    #attrs
+                    const _: fn() = #begin #ident, ::cxx::kind::Trivial #end;
+                });
+            }
 
-    verify
+            verify
+        }
+        Lang::Rust => {
```
dtolnay, line 1296: This is not sufficient for safety. Consider adding:

```rust
        // inside extern "Rust"
        type Any = crate::module::CrossModuleRustType;
        fn repro(bad: &mut CrossModuleRustType) -> &mut Any;
...

fn repro(bad: &mut ffi::CrossModuleRustType) -> &mut ffi::Any {
    bad
}
```

Nothing here enforces what type the C++ `Any` is. We can make it anything:

```cpp
namespace tests {
using Any = std::array<char, 1000>;
}
```

and arbitrarily stomp on memory.

```cpp
repro(*r_boxed_cross_module_rust_type(123)).fill('?');
```

```console
     Running tests/test.rs (target/debug/deps/test-f2f5723622a08a56)

running 15 tests
free(): invalid next size (fast)
malloc(): corrupted top size
error: test failed, to rerun pass `--test test`

Caused by:
  process didn't exit successfully: `target/debug/deps/test-f2f5723622a08a56` (signal: 6, SIGABRT: process abort signal)
```

dtolnay (changes requested):

macro/src/expand.rs
anforowicz: I don't understand why you say that `Any` is a C++ type.  It is a Rust type, right?  (by virtue of `::cxx::private::verify_rust_type` which ensures that `T: RustType` which means the aliased type must have been used in another `cxx::bridge`'s `extern "Rust"` section.

book/src/extern-rust.md
anforowicz: > Reusing an existing defined pair of C++ and Rust type is already supported. I don't understand the distinction between the extern C++ type aliases which are already supported, and the extern Rust type aliases in this PR. What is the situation where it would be correct to write `extern "Rust" { type T = path::to::U; }` and not `extern "C++" { type T = path::to::U; }`?

In the PR, in the end-to-end test under `tests/ffi/...` the `CrossModuleRustType` is a **Rust** type, rather than a C++ type.  This is why `extern "C++" { type T = path::to::U; }` will not work and we need to support `extern "Rust" { type T = path::to::U; }`.  

Let me :

* If `Foo` is a C++ type (declared in another `#[cxx::bridge]`'s `extern "C++"` section) then `extern "C++" { type Foo = crate::some::other::module::Foo; }` will indeed work.
* If `Foo` is a Rust type (declared in another Rust module, and also covered by another `#[cxx::bridge]`) then:
    - `extern "C++" { type Foo = crate::some::other::module::Foo; }` will **not** work, because `Foo` is a Rust type rather than a C++ type/binding.  In particular, if I change [`tests/ffi/lib.rs` in the PR](https://github.com/dtolnay/cxx/pull/1539/commits/8dbb7b28889375ec08c88d1cf04e8670217f83c5#diff-1a4545dcfeac97a87f9a57583433df42197c80da8b3c701dec5aecd6e1ee063b) to put the type alias in an `extern "C++"` section, then `cargo test` rightfully fails saying: "the trait bound `module::CrossModuleRustType: ExternType` is not satisfied".
    - I can't just say `extern "Rust" { type Foo; /* ... */ }` and have `type Foo = crate::some::other::module::Foo` outside of `cxx::bridge`, because this will error out with: "conflicting implementations of trait `RustType` for type `module::CrossModuleRustType`" (or `Foo`, this error message was captured by tweaking the end-to-end tests from the PR).

macro/src/expand.rs
dtolnay: > If the error above (i.e. "free(): invalid next size" and "malloc(): corrupted top size") can be reproed, then I would really appreciate if you could share the repro as a commit on top/after my PR?

This is what I mean: https://github.com/anforowicz/cxx/pull/3. This would need to not compile in order for the implementation to be sound.

anforowicz: Thanks!  That helped.  I understand the problem now.

The _specific_ scenario that you've pointed out should now be fixed in the latest version of the PR.  The fix is to check on C++ side that the type alias is derived from `::rust::Opaque`.  OTOH, the fix doesn't seem sufficient to solve the _general_ problem - transmuting into an unrelated type seems still posslble with something like:

```
#[cxx::bridge]
mod ffi {
    extern "Rust" {
        type RustType;
        type OtherRustType;

        #[cxx_name = "OtherRustType"]
        type RustAlias = RustType;

        fn bad_transmute(x: &RustType) -> &RustAlias;
        fn do_something_with_other_rust_type(x: &OtherRustType);
    }
}
```

To prevent the above, we'd need to assert on C++ side not only that `RustAlias` aliases **an** opaque Rust type, but that it aliases **the** same, specific Rust type.  In other words, this probably requires a C++ equivalent of `ExternType::Id` checks from Rust side.

Q1: Does the above sound more or less correct to you?

Q2: Do you have suggestions on how to implement the C++ equivalent of `ExternType::Id` checks?  Some initial thoughts below:

* It seems that the type id of a Rust type should include the type name, the module path, and the crate name.  The latter seems difficult, because 1) IIUC `cxxbridge` does *not* currently know/get the crate name, and 2) crate names can be aliased.
* One mechanism to add a mechanism for type id checks would be to tweak `write_opaque_type` so that the generated struct has an extra field like so:

```
#include <cstring>

struct S /* : ::rust::Opaque */ {
    static constexpr const char* __cxxbridge1_rust_type_id = "module::path::RustType";
};

static_assert(
    0 == strcmp(S::__cxxbridge1_rust_type_id, "module::path::OtherRustType"),
    "blah"
);
```

(having a new public field is a bit icky but is not the end of the world;  one alternative would be to introduce template parameters into `::rust::Opaque`, but changing this type should probably be treated as a breaking change which seems to make this approach undesirable)

anforowicz: I expected that treating a Rust type as an opaque C++ type will lead to considerable trouble, but so far I can only think of `Box<CrossModuleRustType>` as one specific broken scenario.  `fn check_type_box` explicitly excludes `cx.types.aliases` from its checks, but it seems that `CrossModuleRustType` type still cannot be used as `T` in `Box` or `Vec` - after adding `fn c_boxed_cross_module_rust_type() -> Box<CrossModuleRustType>;` to `extern "C++" { ... }` (or to `extern "Rust" { ... }`) I get:

```
cxxbridge/sources/tests/ffi/lib.rs.cc:1489:58: error: static assertion failed: type tests::CrossModuleRustType should be trivially move constructible and trivially destructible in C++ to be used as type Box<CrossModuleRustType> in Rust
warning: cxx-test-suite@0.0.0:  1489 |     ::rust::IsRelocatable<::tests::CrossModuleRustType>::value,
warning: cxx-test-suite@0.0.0:       |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
```

After tweaking `fn write_opaque_type` to emit `using IsRelocatable = std::true_type;` I can get past the error above, but then another error happens:

    > type mismatch resolving `<CrossModuleRustType as ExternType>::Kind == Trivial`

 This error can be worked around by tweaking `fn expand_rust_type_impl` to emit `type Kind = ::cxx::kind::Trivial` (rather than `Opaque`).  But then this also doesn't work with an error that suggests that `Box<T>` wouldn't work even with my original approach:

    > symbol `tests$cxxbridge1$r_boxed_cross_module_rust_type` is already defined

-------------

Interestingly `&mut T` works fine if I retain the `IsRelocatable` and `kind::Trivial` tweaks.  So maybe this is indeed a better way forward compared to my original approach.

anforowicz:
> I know that problem 2 should go away after my ad-hoc tweaks from [#1539 (comment)](https://github.com/dtolnay/cxx/pull/1539#discussion_r2323702189). I''ll try to spend some time today trying to understand if/why those tweaks are ok/safe (and then putting them into a separate PR and abandoning this PR).

I've submitted https://github.com/dtolnay/cxx/pull/1606 to replace the current PR.

ShabbyX:
> There is also a problem of duplicate monormorphization/impl thunks when Box<RustTypeAlias>, Vec<RustTypeAlias>, etc are used. I hope that this isn't immediately blocking for the ANGLE work

Thanks. I have an ffi::Box<IR> on the C++ side, but if I can pass it to Rust as `&mut IR`, I don't really have a need to take it as `Box<IR>` (i.e. not a blocker)

PR #1606, opened 2025-09-05
Improving support for cross-bridge type aliases of Rust types.

anforowicz:
This PR improves docs (to explain that the type alias has to go into a `extern "C++"` section, and to point out the need to use `#[derive(ExternType)]`).

This commit also fixes errors that would be reported if a type alias for a Rust type would be used as a `&mut T` parameter.  The fixes boil down to ensuring that the Rust type is recognized as Rust-movable.  More details can be found in the PR discussion.

# Description of changes in `macro/src/expand.rs`

https://github.com/dtolnay/cxx/commit/16e2620fe51032d789770b489e27a9c6c27719ef has introduced support for `#[derive(ExternType)]` in `extern “Rust”` section.  I don’t understand why this commit used `type Kind = ::cxx::kind::Opaque` rather than `kind::Trivial`.

Docs for [`cxx::kind::Opaque`](https://docs.rs/cxx/1.0.175/cxx/kind/enum.Opaque.html) say that “an opaque type [...] cannot be passed or held by value within Rust” because “Rust’s move semantics are such that every move is equivalent to a memcpy” (which is “incompatible in general with C++’s constructor-based move semantics”).

Docs for [`cxx::kind::Trivial`](https://docs.rs/cxx/1.0.175/cxx/kind/enum.Trivial.html) say that this is “a type with trivial move constructor and no destructor, which can therefore be owned and moved around in Rust code without requiring indirection”.

Based on [`tests/ui/rust_pinned.rs`](https://github.com/dtolnay/cxx/blob/fff9d28d713a54a1b575b3d798432d2ae8027f0b/tests/ui/rust_pinned.rs) (and the corresponding expected error message in [the `.stderr` file here](https://github.com/dtolnay/cxx/blob/fff9d28d713a54a1b575b3d798432d2ae8027f0b/tests/ui/rust_pinned.stderr) I assume that `cxx` does not allow `extern “Rust” { type RustType; }` for `!Unpin` types.  So this would mean that all Rust types supported by `cxx` can be moved by `memcpy`.

Therefore it seems okay to change `macro/src/[expand.rs](http://expand.rs/)` to emit `type Kind = ::cxx::kind::Trivial`

Without the change in [`macro/src/expand.rs`](https://github.com/dtolnay/cxx/commit/81f774d2b1a44cdf2a600a9c1e8b5e69b78ca4af#diff-e21af94dd924766138d7f7e4602b7cfed26b1976d0c0830d40b9c4b948dd929e), the new tests would result in the following error:

```
error[E0271]: type mismatch resolving `<CrossModuleRustType as ExternType>::Kind == Trivial`
   --> tests/ffi/lib.rs:374:14
    |
374 |         type CrossModuleRustType = crate::module::CrossModuleRustType;
    |              ^^^^^^^^^^^^^^^^^^^ type mismatch resolving `<CrossModuleRustType as ExternType>::Kind == Trivial`
    |
note: expected this to be `Trivial`
   --> tests/ffi/module.rs:85:18
    |
 85 |         #[derive(ExternType)]
    |                  ^^^^^^^^^^
note: required by a bound in `verify_extern_kind`
   --> /usr/local/google/home/lukasza/src/github/cxx/src/extern_type.rs:187:41
    |
187 | pub fn verify_extern_kind<T: ExternType<Kind = Kind>, Kind: self::Kind>() {}
    |                                         ^^^^^^^^^^^ required by this bound in `verify_extern_kind`
```

# Description of changes in `gen/src/write.rs`

Currently C++ bindings for `extern “Rust”` types do **not** mark the bindings as relocatable (in the [`cxx::IsRelocatable<T>`](https://github.com/dtolnay/cxx/blob/fff9d28d713a54a1b575b3d798432d2ae8027f0b/include/cxx.h#L489-L490) sense).

The doc comment of `cxx::IsRelocatable` says this is an assertion that the type “is soundly relocatable by Rust”.

We already established above that all Rust types supported by `cxx` can be moved by `memcpy`.  Therefore from that perspective it seems okay to start emitting `using IsRelocatable = ::std::true_type` inside the struct/bindings emitted for Rust types by `fn write_opaque_type` in `gen/src/[write.rs](http://write.rs/)`.

Emitting a new public member carries a risk that its name may conflict/clash with another, preexisting member.  So from that perspective, it may be safer to directly specialize `cxx::IsRelocatable`.  OTOH, [the Rust naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) reserve UpperCamelCase to types, traits, enum variants, and type parameters - this means that a clash with `IsRelocatable` seems unlikely.

Without the change in [`gen/src/write.rs`](https://github.com/dtolnay/cxx/commit/81f774d2b1a44cdf2a600a9c1e8b5e69b78ca4af#diff-7719228d0694503ed3ea2dd14d6561e3bcaad827e0ce3c8f2433b70476dff099), the new tests would result in the following error:

```
warning: cxx-test-suite@0.0.0: /usr/local/google/home/lukasza/src/github/cxx/target/debug/build/cxx-test-suite-690332097c2cd1c3/out/cxxbridge/sources/tests/ffi/lib.rs.cc:1489:58: error: static assertion failed: type tests::CrossModuleRustType should be trivially move constructible and trivially destructible in C++ to be used as a non-pinned mutable reference in signature of `r_mut_ref_cross_module_rust_type` in Rust
warning: cxx-test-suite@0.0.0:  1489 |     ::rust::IsRelocatable<::tests::CrossModuleRustType>::value,
warning: cxx-test-suite@0.0.0:       |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
error: failed to run custom build command for `cxx-test-suite v0.0.0 (/usr/local/google/home/lukasza/src/github/cxx/tests/ffi)`
```

----------------

/cc @ShabbyX

### DIFF
--- a/book/src/extern-c++.md
+++ b/book/src/extern-c++.md
@@ -189,23 +189,40 @@ manually implemented as described below.
 
 #### Safely unifying occurrences of an extern type across bridges
 
-In the following snippet, two #\[cxx::bridge\] invocations in different files
+In the following snippet, two `#[cxx::bridge]` invocations in different files
 (possibly different crates) both contain function signatures involving the same
-C++ type `example::Demo`. If both were written just containing `type Demo;`,
+C++ type `example::CppType`. If both were written just containing
+`type CppType;`,
 then both macro expansions would produce their own separate Rust type called
-`Demo` and thus the compiler wouldn't allow us to take the `Demo` returned by
-`file1::ffi::create_demo` and pass it as the `Demo` argument accepted by
-`file2::ffi::take_ref_demo`. Instead, one of the two `Demo`s has been defined as
+`CppType` and thus the compiler wouldn't allow us to take the `CppType` returned
+by
+`file1::ffi::create_cpp_type` and pass it as the `CppType` argument accepted by
+`file2::ffi::take_cpp_type_by_ref`. Instead, one of the two `CppType`s has been
+defined as
 an extern type alias of the other, making them the same type in Rust.
+
+Problems would also happen if `type RustType` appeared in both `#[cxx::bridge]`
+invocations.  The solution is similar - replacing one of the occurences with
+an extern type alias of the other type.  Note that this type alias has to be
+present in `extern "C++"` section even though we are aliasing a Rust type.
+Also note that the type alias depends on `impl ExternType for RustType` which
+the example below generates via `#[derive(ExternType)]`.
 
 ```rust,noplayground
 // file1.rs
 #[cxx::bridge(namespace = "example")]
 pub mod ffi {
     unsafe extern "C++" {
-        type Demo;
+        type CppType;
 
-        fn create_demo() -> UniquePtr<Demo>;
+        fn create_cpp_type() -> UniquePtr<CppType>;
+    }
+
+    extern "Rust {
+        #[derive(ExternType)]
+        type RustType;
+
+        fn create_rust_type() -> Box<RustType>;
     }
 }
 ```
@@ -215,9 +232,14 @@ pub mod ffi {
 #[cxx::bridge(namespace = "example")]
 pub mod ffi {
     unsafe extern "C++" {
-        type Demo = crate::file1::ffi::Demo;
+        type CppType = crate::file1::ffi::CppType;
+        type RustType = crate::file1::ffi::RustType;
 
-        fn take_ref_demo(demo: &Demo);
+        fn take_cpp_type_by_ref(x: &CppType);
+    }
+
+    extern "Rust {
+        fn take_rust_type_by_ref(x: &RustType);
     }
 }
 ```
--- a/gen/src/write.rs
+++ b/gen/src/write.rs
@@ -427,6 +427,7 @@ fn write_opaque_type<'a>(out: &mut OutFile<'a>, ety: &'a ExternType, methods: &[
     }
 
     writeln!(out, "  ~{}() = delete;", ety.name.cxx);
+    writeln!(out, "  using IsRelocatable = std::true_type;");
     writeln!(out);
 
     out.builtin.layout = true;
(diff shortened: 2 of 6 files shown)


dtolnay (changes requested):
This would allow exposing an ABI mismatch where types in a struct or function signature have different offsets in different languages, which is unsound. `cxx::kind::Opaque` is required if the Rust and C++ definition are differently sized or aligned.

Example:

```rust
// src/module.rs

pub struct Rust(pub [isize; 2]);

#[cxx::bridge]
pub mod ffi {
    extern "Rust" {
        #[derive(ExternType)]
        type Rust;
    }
}
```

```rust
// src/main.rs

pub mod module;

#[cxx::bridge]
mod ffi {
    extern "C++" {
        include!("example/src/module.rs.h");

        type Rust = crate::module::Rust;
    }

    unsafe extern "C++" {
        fn repro(s: &Struct);
    }

    struct Struct {
        rust: Rust,
        string: String,
    }
}

fn main() {
    let s = ffi::Struct {
        rust: module::Rust([1; 2]),
        string: ".".repeat(100),
    };
    println!("{:p}", &s);
    println!("{:p}", &s.string);
    ffi::repro(&s);
}
```

```cpp
// include/header.h

#pragma once
struct Struct;
void repro(Struct const &s) noexcept;
```

```cpp
// src/lib.cc

#include <example/src/main.rs.h>
#include <rust/cxx.h>
#include <iostream>

void repro(Struct const &s) noexcept {
  std::cout << static_cast<const void*>(&s);
  std::cout << '\n';
  std::cout << static_cast<const void*>(&s.string);
  std::cout << '\n';
  std::cout << s.string;
  std::cout << '\n';
}
```

```console
0x7ffc54784ee0
0x7ffc54784ef0
0x7ffc54784ee0
0x7ffc54784ee8
Segmentation fault (core dumped)
```

PR #1658, opened 2025-09-29
Arbitrary inner type in named impl key

anforowicz:
PTAL?

If this PR looks okay, then the next commit/PR (https://github.com/anforowicz/cxx/commit/8ab4eccedec87c7a913b140dfd589b1af2cebb82) should be able to add support for `Vec<Box<T>>` (because of how `Box<T>` is [special](https://doc.rust-lang.org/reference/items/implementations.html#r-items.impl.trait.fundamental) when evaluating the orphan rule.

/cc @zetafunction who has kindly provided initial feedback/review at  https://github.com/anforowicz/cxx/pull/4

### DIFF
--- a/macro/src/expand.rs
+++ b/macro/src/expand.rs
@@ -1656,19 +1656,19 @@ fn expand_rust_box(
     types: &Types,
     conditional_impl: &ConditionalImpl,
 ) -> TokenStream {
-    let ident = key.rust;
-    let resolve = types.resolve(ident);
-    let link_prefix = format!("cxxbridge1$box${}$", resolve.name.to_symbol());
+    let inner = key.inner;
+    let link_prefix = format!("cxxbridge1$box${}$", key.symbol);
     let link_alloc = format!("{}alloc", link_prefix);
     let link_dealloc = format!("{}dealloc", link_prefix);
     let link_drop = format!("{}drop", link_prefix);
 
-    let local_prefix = format_ident!("{}__box_", ident);
+    let local_prefix = format_ident!("{}__box_", key.symbol);
     let local_alloc = format_ident!("{}alloc", local_prefix);
     let local_dealloc = format_ident!("{}dealloc", local_prefix);
     let local_drop = format_ident!("{}drop", local_prefix);
 
-    let (impl_generics, ty_generics) = generics::split_for_impl(key, conditional_impl, resolve);
+    let (impl_generics, ty_generics) =
+        generics::get_impl_and_ty_generics(inner, conditional_impl, types);
 
     let cfg = conditional_impl.cfg.into_attr();
     let begin_span = conditional_impl
@@ -1678,18 +1678,18 @@ fn expand_rust_box(
         .explicit_impl
         .map_or(key.end_span, |explicit| explicit.brace_token.span.join());
     let unsafe_token = format_ident!("unsafe", span = begin_span);
-    let prevent_unwind_drop_label = format!("::{} as Drop>::drop", ident);
+    let prevent_unwind_drop_label = quote! { #inner }.to_string();
 
     quote_spanned! {end_span=>
         #cfg
         #[automatically_derived]
         #[doc(hidden)]
-        #unsafe_token impl #impl_generics ::cxx::private::ImplBox for #ident #ty_generics {}
+        #unsafe_token impl #impl_generics ::cxx::private::ImplBox for #inner #ty_generics {}
 
         #cfg
         #[doc(hidden)]
         #[unsafe(export_name = #link_alloc)]
-        unsafe extern "C" fn #local_alloc #impl_generics() -> *mut ::cxx::core::mem::MaybeUninit<#ident #ty_generics> {
+        unsafe extern "C" fn #local_alloc #impl_generics() -> *mut ::cxx::core::mem::MaybeUninit<#inner #ty_generics> {
             // No prevent_unwind: the global allocator is not allowed to panic.
             //
             // TODO: replace with Box::new_uninit when stable.
@@ -1701,7 +1701,7 @@ fn expand_rust_box(
         #cfg
         #[doc(hidden)]
         #[unsafe(export_name = #link_dealloc)]
-        unsafe extern "C" fn #local_dealloc #impl_generics(ptr: *mut ::cxx::core::mem::MaybeUninit<#ident #ty_generics>) {
+        unsafe extern "C" fn #local_dealloc #impl_generics(ptr: *mut ::cxx::core::mem::MaybeUninit<#inner #ty_generics>) {
             // No prevent_unwind: the global allocator is not allowed to panic.
             let _ = unsafe { ::cxx::alloc::boxed::Box::from_raw(ptr) };
         }
@@ -1709,7 +1709,7 @@ fn expand_rust_box(
         #cfg
         #[doc(hidden)]
         #[unsafe(export_name = #link_drop)]
-        unsafe extern "C" fn #local_drop #impl_generics(this: *mut ::cxx::alloc::boxed::Box<#ident #ty_generics>) {
+        unsafe extern "C" fn #local_drop #impl_generics(this: *mut ::cxx::alloc::boxed::Box<#inner #ty_generics>) {
             let __fn = ::cxx::core::concat!("<", ::cxx::core::module_path!(), #prevent_unwind_drop_label);
             ::cxx::private::prevent_unwind(__fn, || unsafe { ::cxx::core::ptr::drop_in_place(this) });
         }
@@ -1721,9 +1721,8 @@ fn expand_rust_vec(
     types: &Types,
     conditional_impl: &ConditionalImpl,
 ) -> TokenStream {
-    let elem = key.rust;
-    let resolve = types.resolve(elem);
-    let link_prefix = format!("cxxbridge1$rust_vec${}$", resolve.name.to_symbol());
+    let inner = key.inner;
+    let link_prefix = format!("cxxbridge1$rust_vec${}$", key.symbol);
     let link_new = format!("{}new", link_prefix);
     let link_drop = format!("{}drop", link_prefix);
     let link_len = format!("{}len", link_prefix);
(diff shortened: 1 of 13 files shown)


syntax/mangle.rs
```diff
@@ -118,3 +118,25 @@ pub(crate) fn c_trampoline(efn: &ExternFn, var: &Pair, types: &Types) -> Symbol
...
+/// that can be used as a **part** of monomorphized/instantiated thunk names
+/// (e.g. `cxxbridge1$box$org$rust$Struct$alloc`).
+///
+/// Not all type names can be mangled at this point - `None` will be returned if
+/// mangling fails.  We have to gracefully handle non-manglable types, because
+/// some callers (e.g. `Type`'s `impl_key` method) call into `mangle::type_`
+/// before `syntax/check.rs` has rejected unsupported generic type parameters.
+pub(crate) fn type_(t: &Type) -> Option<Symbol> {
+    match t {
+        Type::Ident(named_type) => Some(join!(named_type.rust)),
```
dtolnay: This is going to give conflicting symbols when an identical Rust name exists in multiple namespaces. The original code before this PR was mangling using the C++ namespace and name.

```rust
// src/lib.rs

mod two;

#[cxx::bridge(namespace = "one")]
pub mod one {
    struct Context { x: i32 }
    impl Box<Context> {}
}
```

```rust
// src/two.rs

#[cxx::bridge(namespace = "two")]
pub mod two {
    struct Context { x: i8 }
    impl Box<Context> {}
}
```

```console
error: symbol `cxxbridge1$box$Context$alloc` is already defined
 --> src/two.rs:6:23
  |
6 |     impl Box<Context> {}
  |                       ^^
```

or for a C++ type:

```console
  = note: rust-lld: error: duplicate symbol: cxxbridge1$unique_ptr$Context$null
          >>> defined at main.rs.cc:64 (target/debug/build/example-aff44aa21d738c2a/out/cxxbridge/sources/example/src/main.rs.cc:64)
          >>>            4be10217b215df70-main.rs.o:(cxxbridge1$unique_ptr$Context$null) in archive target/debug/build/example-aff44aa21d738c2a/out/libexample-cxx.a
          >>> defined at two.rs.cc:64 (target/debug/build/example-aff44aa21d738c2a/out/cxxbridge/sources/example/src/two.rs.cc:64)
          >>>            4be10217b215df70-two.rs.o:(.text.cxxbridge1$unique_ptr$Context$null+0x0) in archive target/debug/build/example-aff44aa21d738c2a/out/libexample-cxx.a
```

macro/src/expand.rs
```diff
@@ -1678,18 +1678,18 @@ fn expand_rust_box(
         .explicit_impl
         .map_or(key.end_span, |explicit| explicit.brace_token.span.join());
     let unsafe_token = format_ident!("unsafe", span = begin_span);
-    let prevent_unwind_drop_label = format!("::{} as Drop>::drop", ident);
+    let prevent_unwind_drop_label = quote! { #inner }.to_string();
```
dtolnay: This is not an adequate replacement for what was here before.

```cpp
#pragma once
#include <rust/cxx.h>

class Thing;

inline void take_thing(rust::Box<Thing>) {}
```

```rust
#[cxx::bridge]
mod ffi {
    extern "Rust" {
        type Thing;
    }
    unsafe extern "C++" {
        fn take_thing(_: Box<Thing>);
    }
}

struct Thing;

impl Drop for Thing {
    fn drop(&mut self) {
        panic!("...");
    }
}

fn main() {
    ffi::take_thing(Box::new(Thing));
}
```

Before:

```console
panic in ffi function <example::ffi::Thing as Drop>::drop, aborting.
```

After:

```console
panic in ffi function <example::ffiThing, aborting.
```

syntax/names.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) struct ForeignName {
 }
 
 impl Pair {
+    #[allow(dead_code)] // only used by cxx-gen, not cxxbridge-macro
```
dtolnay:
```suggestion
    #[cfg_attr(proc_macro, expect(dead_code))]
```

syntax/instantiate.rs
```diff
@@ -13,71 +13,95 @@ pub(crate) enum ImplKey<'a> {
...
-    pub rust: &'a Ident,
-    #[cfg_attr(not(proc_macro), expect(dead_code))]
-    pub lt_token: Option<Token![<]>,
-    #[cfg_attr(not(proc_macro), expect(dead_code))]
-    pub gt_token: Option<Token![>]>,
-    #[cfg_attr(not(proc_macro), expect(dead_code))]
+    /// Mangled form of the `outer` type.
+    pub symbol: Symbol,
+    /// Generic type - e.g. `UniquePtr<u8>`.
+    #[allow(dead_code)] // only used by cxx-build, not cxxbridge-macro
```
dtolnay:
```suggestion
    #[cfg_attr(proc_macro, expect(dead_code))]
```

syntax/instantiate.rs
```diff
@@ -13,71 +13,95 @@ pub(crate) enum ImplKey<'a> {
...
+    pub outer: &'a Type,
+    /// Generic type argument - e.g. `u8` from `UniquePtr<u8>`.
+    pub inner: &'a Type,
+    #[allow(dead_code)] // only used by cxxbridge-macro, not cxx-build
```
dtolnay:
```suggestion
    #[cfg_attr(not(proc_macro), expect(dead_code))]
```

macro/src/generics.rs
```diff
@@ -1,65 +1,85 @@
...
+            //             type Borrowed<'a>;
+            //             fn borrowed(arg: &i32) -> UniquePtr<Borrowed>;
+            //         }
+            //     }
+            //
+            // But this could have also been spelled with `explicit_generics`:
+            //
+            //             fn borrowed<'a>(arg: &'a i32) -> UniquePtr<Borrowed<'a>>;
+            let explicit_generics = get_generic_lifetimes(inner);
+            if explicit_generics.lifetimes.is_empty() {
```
dtolnay: This is incorrect in the case of `<>`. For example this compiles successfully before this PR:

```rust
#[cxx::bridge]
mod ffi {
    struct Borrowed<'a> {
        s: &'a str,
    }
    extern "Rust" {
        fn f(_: UniquePtr<Borrowed<>>);
    }
}

fn f(_: cxx::UniquePtr<ffi::Borrowed<>>) {}
```

but generates syntactically invalid code after:

```console
error: expected one of `!`, `+`, `::`, `where`, or `{`, found `<`
 --> src/main.rs:3:20
  |
3 |     struct Borrowed<'a> {
  |                    ^ unexpected token
...
7 |         fn f(_: UniquePtr<Borrowed<>>);
  |                                     - expected one of `!`, `+`, `::`, `where`, or `{`
```

macro/src/expand.rs
```diff
@@ -1985,14 +1983,13 @@ fn expand_shared_ptr(
...
-    let not_destructible_err = format!("{} is not destructible", display_namespaced(resolve.name));
 
     quote_spanned! {end_span=>
         #cfg
         #[automatically_derived]
-        #unsafe_token impl #impl_generics ::cxx::memory::SharedPtrTarget for #ident #ty_generics {
+        #unsafe_token impl #impl_generics ::cxx::memory::SharedPtrTarget for #inner #ty_generics {
             fn __typename(f: &mut ::cxx::core::fmt::Formatter<'_>) -> ::cxx::core::fmt::Result {
-                f.write_str(#name)
+                f.write_str(::core::stringify!(#inner))
```
dtolnay: Using stringify for this is not a good approach.

```rust
#[cxx::bridge]
pub mod ffi {
    struct Context<'a> {
        session: &'a str,
    }
    struct Server<'srv> {
        ctx: SharedPtr<Context<'srv>>,
    }
    struct Client<'clt> {
        ctx: SharedPtr<Context<'clt>>,
    }
}

fn main() {
    let _ = *cxx::SharedPtr::<ffi::Context>::null();
}
```

Before:

```console
called deref on a null SharedPtr<Context>
```

After:

```console
called deref on a null SharedPtr<Context < 'srv >>
```

The spacing is wonky, and this is using a random one of the lifetime names which is going to be misleading for the user if their panic is happening in "client" code that has nothing to do with "server" code.

macro/src/generics.rs
```diff
@@ -1,65 +1,85 @@
...
+                // In the example above, we want to use generics from `type Borrowed<'a>`.
+                let resolved_generics = resolve_generic_lifetimes(inner, types);
+                (resolved_generics, Some(resolved_generics))
+            } else {
+                (
+                    explicit_generics,
```
dtolnay: This `explicit_generics` is never the best set of lifetime names to use for an impl. In this codepath there is no explicit impl, so `explicit_generics` is populated with lifetime names from some random usage of the generic type in a function signature or struct. In general those lifetime names are not meaningful at all outside the context of that single function or struct. This code should not be propagating them onto an impl block (which is going to become visible in rustdoc).

For example:

```rust
struct Context<'sess> {
    session: &'sess str,
}
struct Server<'srv> {
    ctx: SharedPtr<Context<'srv>>,
}
struct Client<'clt> {
    ctx: SharedPtr<Context<'clt>>,
}
```

It does not make sense for the documentation of `Context` to be rendered as:

- `impl<'sess> ExternType for Context<'sess>`
- `impl<'srv> SharedPtrTarget for Context<'srv>`

The correct impl would be `impl<'sess> SharedPtrTarget for Context<'sess>` (which was generated before this PR).

syntax/mangle.rs
```diff
@@ -118,3 +118,25 @@ pub(crate) fn c_trampoline(efn: &ExternFn, var: &Pair, types: &Types) -> Symbol
...
+        Type::RustBox(ty1) => type_(&ty1.inner).map(|s| join!("box", s)),
+        Type::RustVec(ty1) => type_(&ty1.inner).map(|s| join!("rust_vec", s)),
+        Type::UniquePtr(ty1) => type_(&ty1.inner).map(|s| join!("unique_ptr", s)),
+        Type::SharedPtr(ty1) => type_(&ty1.inner).map(|s| join!("shared_ptr", s)),
+        Type::WeakPtr(ty1) => type_(&ty1.inner).map(|s| join!("weak_ptr", s)),
```
dtolnay: These are all currently unreachable, right?
```suggestion
```

dtolnay (commented):
Thank you &mdash; I am merging this with some changes in #1669.

syntax/mangle.rs
anforowicz: Yes, correct.

Issue #1507, opened 2025-04-18
Compiler-enforce CXX version matching between C++ and Rust


nyurik:
As discovered in #1506, it is relatively easy to have a mismatch between the cxx-bridge and the cxx crate in a complex project - i.e. cxx-bridge of one version generates the C++ code, and cxx crate of another version is built as part of Rust code.  As anything else with Rust, I think this undefined behavior should be caught by the compiler if possible.  FFI tends to be super error-prone, so reducing error surface would help.

nyurik:
I have been thinking of the ways to enforce it, but if I understand it correctly, the bridge (c++) and crate (rust) exist on different sides of the language divide, which means the only way to catch this is via some linker magic.  For example, given this declaration:

```rust
#[cxx::bridge(namespace = "rustutils")]
mod ffi { /*...*/ }
```

the cxx bridge would generate a dummy function, and the `cxx::bridge` rust macro would generate a call to that function

```
namespace rustutils {
  void $$cxx_bridge_version_1_0_157() {}
}
```

The function name should be descriptive enough so that if a linker shows missing symbol, it would be somewhat easy for developer to understand the problem. A similar approach could be used to verify the reverse - macro declaring a function and C++ calling it (I am not certain both are needed).

Lastly, I wonder if this approach would result in a (small) binary size increase, or if a linker could eliminate the empty functions. This should not really matter for the regular apps, but the embedded folks tend to worry about the binary size.

sdht0:
Just got bit by this mismatch issue as well.

PR #1665, opened 2025-10-27
Encode patch-version in the prefix of cxx-generated symbols.

anforowicz:
Fixes https://github.com/dtolnay/cxx/issues/1507

### DIFF
--- a/syntax/mangle.rs
+++ b/syntax/mangle.rs
@@ -7,6 +7,7 @@
 //          defining characteristics:
 //             - 2 segments
 //             - starts with cxxbridge
+//          TODO: should these also include {CXXVERSION}?
 //
 //   (b) Behavior on a builtin binding without generic parameter.
 //          pattern:  {CXXBRIDGE} $ {TYPE} $ {NAME}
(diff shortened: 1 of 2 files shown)


anforowicz:
> should the symbol name be something more descriptive to explain to the user what they are doing wrong?

I don't think so - the prefix will be present in each `cxx`-generated symbol.  If the prefix is too long, then the symbols may get unwieldy to use (e.g. when reading them on a limited-width terminal, or when writing them to set breakpoints or to describe bugs).

strike-ntzachar:
> > a way of exposing unmangled symbols directly from my shared library without requiring linking to any glue code
> 
> Would [the `cxx_gen::Opt::cxx_impl_annotations`](https://docs.rs/cxx-gen/latest/cxx_gen/struct.Opt.html#structfield.cxx_impl_annotations) option help in your scenario without requiring manually-authored glue code with hardcoded mangled names?

As far as I understand this option, no.
I'll try to explain further. Say I have the following:
```rust
#[cxx::bridge(namespace = "alpine")]
pub mod ffi {
   extern "Rust" {
      fn init() -> i32;
   }
}
```
I then compile my crate as cdynlib. This results in a shared library, with the following symbol exported: `alpine$cxxbridge1$190$init`
As far as I understand, this option will not help (not to mention that I have no idea how to use it...)

nyurik:
The above discussion made my day...

image

Issue #1676, opened 2025-11-30
Symbol mismatch with cxxbridge-cmd in Bazel


sayrer:
In my build, the linker was looking for symbols like `_twitter_text$cxxbridge1$0$autolink_all` but the Rust static library contained symbols like `_twitter_text$cxxbridge1$189$autolink_all`. This happened because:

1. The `cxxbridge-cmd` tool generates C++ bridge code with version `$0$` when processing standalone `.rs` files
2. The Rust compiler embeds the actual cxx crate version (1.0.189, hence `$189$`) into symbols when building the library

am009:
I got the same issue.

In my project, I have:

```
[dependencies]
cxx = { version = "1.0", default-features = false, features = ["alloc"] }
```

Then I got the error:

```
rust-lld: error: undefined symbol: cxxbridge1$coverage_new
```

If I changed to specify the specific patch version

```
[dependencies]
cxx = { version = "1.0.190", default-features = false, features = ["alloc"] }
```

It seems to work without error.

sayrer:
OK, this works now. Maybe the docs use this as an example? It's pretty sparse right now.

```
#
# The codegen tool needed by cxx.
#

CXX_BRIDGE_VERSION = "1.0.190"

http_archive(
    name = "cxxbridge-cmd",
    build_file_content = """
load("@rules_rust//rust:defs.bzl", "rust_binary")
load("@cxxbridge_cmd_deps//:defs.bzl", "aliases", "all_crate_deps")

rust_binary(
    name = "cxxbridge-cmd",
    srcs = glob(["src/**/*.rs"]),
    aliases = aliases(),
    compile_data = glob(["src/gen/**/*.h"]),
    edition = "2021",
    visibility = ["//visibility:public"],
    deps = all_crate_deps(
        normal = True,
    ),
    version = "%s",
)
    """ % CXX_BRIDGE_VERSION,
    sha256 = "b1f29a879d35f7906e3c9b77d7a1005a6a0787d330c09dfe4ffb5f617728cb44",
    strip_prefix = "cxxbridge-cmd-" + CXX_BRIDGE_VERSION,
    type = "tar.gz",
    urls = ["https://static.crates.io/crates/cxxbridge-cmd/cxxbridge-cmd-" + CXX_BRIDGE_VERSION + ".crate"],
)

cxxbridge_cmd_deps = use_extension("@rules_rust//crate_universe:extensions.bzl", "crate")
[...]

PR #1679, opened 2025-12-04
Cover top-level `BUILD.bazel` + `MODULE.bazel` in mdbook Bazel chapter.

anforowicz:
PTAL?

The mdbook `{{#include ...}}` syntax is based on what I found in https://rust-lang.github.io/mdBook/format/mdbook.html#including-portions-of-a-file.  I tested manually by running `DISPLAY=:20 mdbook build -o` under `book/` and looking at the Bazel chapter.
 
I don't really use Bazel often, but I hope that including a more complete example helps.  Not sure if a more canonical/reusable example is possible - I hope that Bazel experts if any can chime in with additional/future PRs if needed.

I understand that this PR probably doesn't directly help the reporter of https://github.com/dtolnay/cxx/issues/1676 (since they have more of a custom setup with an ad-hoc `rust_binary` target for `cxxbridge-cmd` and with `cxxbridge-macro` defined somewhere else (hopefully with a matching version).

### DIFF
--- a/book/src/build/bazel.md
+++ b/book/src/build/bazel.md
@@ -35,84 +35,23 @@ as an illustration of one possible working pattern.
 ```python
 # tools/bazel/rust_cxx_bridge.bzl
 
-load("@bazel_skylib//rules:run_binary.bzl", "run_binary")
-load("@rules_cc//cc:defs.bzl", "cc_library")
-
-def rust_cxx_bridge(name, src, deps = []):
-    native.alias(
-        name = "%s/header" % name,
-        actual = src + ".h",
-    )
-
-    native.alias(
-        name = "%s/source" % name,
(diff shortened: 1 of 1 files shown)


sayrer:
> I understand that this PR probably doesn't directly help the reporter of #1676 (since they have more of a custom setup with an ad-hoc `rust_binary` target for `cxxbridge-cmd` and with `cxxbridge-macro` defined somewhere else (hopefully with a matching version).

It all matches, but the [Bazel CXX docs](https://cxx.rs/build/bazel.html) leave this part unexplained:

```
        tool = "//:codegen",
```

I wish the cxxbridge command came as a separate target alongside "cxx" and "cxx_cc", which all works fine. Then, the version numbers would always match.

Issue #1222, opened 2023-06-01
unsupported element type of Vec


OlivierLDff:
# Context

I'm beginner with the library, so I'm relying a lot on the book. I read the book about the opaque rust type that I should use a `Box` to expose that to c++.

# What I wanted to do

- I created a first function `new_my_type() -> Box<MyType>`.
- Then I wanted to add a `fetch_some_types -> Vec<Box<MyType>>` and I got `unsupported element type of Vec`.

# How i fixed it

- of course  `fetch_some_types -> Vec<MyType>` was enough for it to work. But I couldn't find example in the book.

# What could be improved

It would be nice to have some more documentation in the book about `Vec`, that type should be wrapped in `struct` for it to work. I've found the solution with https://github.com/dtolnay/cxx/issues/671.

We could add in the book categories:
- Vec of opaque rust type (ie you don't need to put a Box
- Vec of an unsupported element: wrap it in a struct.

PR #1681, opened 2025-12-11
Add support for `Vec<Box<T>>`.

anforowicz:
PTAL?

This PR fixes https://github.com/dtolnay/cxx/issues/1222.  I've split this into two separate commits to get some minor refactoring out of the way in the 1st comment (see the commit description for more details).

### DIFF
--- a/syntax/check.rs
+++ b/syntax/check.rs
@@ -133,6 +133,10 @@ fn check_type_rust_vec(cx: &mut Check, ty: &Ty1) {
             }
         }
         Type::Str(_) => return,
+        Type::RustBox(ty1) => {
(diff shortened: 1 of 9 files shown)


syntax/check.rs
dtolnay: Explicit impls also need to allow this.

```console
error: unsupported Self type of explicit impl
 --> src/lib.rs:8:5
  |
8 |     impl Vec<Box<Thing>> {}
  |     ^^^^^^^^^^^^^^^^^^^^^^^
```

dtolnay (changes requested):
Thanks!

PR #1682, opened 2025-12-11
Use `Box::new_uninit` instead of `Box::new(MaybeUninit::uninit())`.

anforowicz:
This is a minor refactoring that addresses a TODO that I recently came across.

*  [`Box::new_uninit`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.new_uninit) has been stabilized in 1.82.0
* `cxx` documents in the `README.md` file that it "requires rustc 1.82+"


dtolnay (approved):
Thanks!

PR #1690, opened 2025-12-29
Use a portable isize definition.

Prabhuk:
Remove platform specific headers which were included for ssize_t and use intptr_t instead to prevent failures in baremetal builds.

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -14,7 +14,6 @@
 #include <type_traits>
 #include <utility>
 #include <vector>
-#endif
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Do you have an idea how to preserve this assertion in the cases that ssize_t is available?

https://github.com/dtolnay/cxx/blob/cd508fc309bb61d579d8d5e0f9192dc4fc25cf88/src/cxx.cc#L445-L448

PR #1724, opened 2026-06-29
Add string_view constructor to rust::Str

matts1:
This is much safer than manually providing pointers and size.

Note: a few lines later, we already have `operator std::string_view` defined. So it makes sense to have it in the other direction too.


anforowicz:
This seems like a potential duplicate of https://github.com/dtolnay/cxx/pull/1701?

matts1:
Looks like it, happy to close this PR if the other PR is getting landed

PR #1728, opened 2026-07-03
Make RustCopy accept mutable, temporary, and Array inputs

danlapid:
`kjObject.as<RustCopy>()` only bound `kj::ArrayPtr<const T>&` -- a non-const lvalue reference. Template deduction therefore rejected a mutable `kj::ArrayPtr<T>`, the reference could not bind a temporary (so `arr.asBytes().as<RustCopy>()` failed), and there was no overload for `kj::Array<T>` at all. Callers had to hand-bind an intermediate `kj::ArrayPtr<const T>` lvalue before every owned copy.

Take the `ArrayPtr` by value instead: an `ArrayPtr` is just a pointer+size, so the copy is free, and by-value binds const/mutable, lvalues, and temporaries while still deducing `T`. `kj::Decay` strips the element's constness so `kj::ArrayPtr<const T>` still yields a `rust::Vec<T>`. Add a `kj::Array<T>` overload so arrays convert directly without a manual `.asPtr()`.

Backward compatible: existing `ArrayPtr<const T>` lvalue callers and the `StringPtr`/`ConstString` delegators deduce to the same `rust::Vec<T>`. Adds convert-test coverage for the mutable, temporary, and `kj::Array` cases.


danlapid:
Sorry, wrong upstream

PR #1748, opened 2026-08-10
Add monospace tags and qualify Rust types outside of prelude

mikomikotaishi:
The tables are missing monospace tags for the type names, which renders them as regular text rather than as monospaced text; this PR adds the code tags around them.

Additionally, the names outside of the Rust prelude are fully qualified, to be consistent with the C++ names which are fully qualified, however I did not qualify the names in the prelude.

Finally, I replaced nested namespace declarations (`namespace rust { namespace behavior { /* ... */ } }`) with concise namespace declarations (`namespace rust::behavior { /* ... */ }`) in documentation, which is clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,43 +76,43 @@ supports a `put` operation for a discontiguous buffer upload. For example we
 might be uploading snapshots of a circular buffer which would tend to consist of
 2 chunks, or fragments of a file spread across memory for some other reason.
 
-A runnable version of this example is provided under the *demo* directory of
+A runnable version of this example is provided under the *`demo`* directory of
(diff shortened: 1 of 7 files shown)


mikomikotaishi:
@dtolnay any chance you could look at this? It's only a few documentation/formatting changes

dtolnay:
I have taken the `namespace` changes in 8ec18748a7a79af1618c5e863cfb96cdfc28e9f7. The rest of this is fine as-is, I think.

Thanks!

mikomikotaishi:
While I don't mean any disrespect, is there a reason you don't want to take the monospace formatting changes, especially for things like the comparison table?

PR #1754, opened 2026-09-04
Do not classify references as guaranteed POD

### LINKED ISSUE #1753: `-Wreturn-type-c-linkage` when returning a non-trivial struct
# Repro steps

```
$ git diff origin/master
...
diff --git a/tests/cpp_ui_tests.rs b/tests/cpp_ui_tests.rs
index 06a8efd9..02c6d14d 100644
--- a/tests/cpp_ui_tests.rs
+++ b/tests/cpp_ui_tests.rs
@@ -26,3 +26,44 @@ fn test_unique_ptr_of_incomplete_foward_declared_pointee() {
     let err_msg = test.compile().expect_single_error();
     assert!(err_msg.contains("definition of `::ForwardDeclaredType` is required"));
 }
+
+/// This is a regression test for returning a struct with reference members across
+/// extern "C" thunks without triggering Clang's `-Wreturn-type-c-linkage`.
+///
+/// Note that the original repro required explicitly opting into extra C/C++
+/// warnings (as already done by `.github/workflows/ci.yml`):
+///
+/// ```sh
+/// $ CXX=clang++ CXXFLAGS="-Werror -Wall -Wpedantic" cargo test --test test
+/// ```
+///
+/// This is a separate test from `tests/ffi/lib.rs` and `tests/ffi/module.rs`
+/// to avoid the risk that some other input (e.g. callback functions) will
+/// suppress the warning via `out.pragma.return_type_c_linkage = true` and
+/// `#pragma clang diagnostic ignored "-Wreturn-type-c-linkage"`.
+#[test]
+fn test_return_struct_with_ref() {
+    let test = cpp_compile::Test::new(quote! {
+        #[cxx::bridge]
+        mod ffi {
+            struct StructWithRef<'a> {
+                r: &'a usize,
+            }
+
+            unsafe extern "C++" {
+                include!("include.h");
+                fn c_return_struct_with_ref<'a>(r: &'a usize) -> StructWithRef<'a>;
+            }
+        }
+    });
+    test.write_file(
+        "include.h",
+        indoc! {"
+            #pragma once
+            #include <cstddef>
+            struct StructWithRef;
+            StructWithRef c_return_struct_with_ref(size_t const &r);
+        "},
+    );
+    test.compile().assert_success();
+}

$ CXX=clang++ CXXFLAGS="-Werror -Wall -Wpedantic" cargo test --test cpp_ui_tests
...
test test_return_struct_with_ref ... FAILED
...
.../src/github/cxx/target/debug/build/scratch/54a15552ce707c7d/out/cxx-test-suite/cpp_ui_tests-8830oH/cxx_bridge.generated.cc:24:17: error: 'cxxbridge1$199$c_return_struct_with_ref' has C-linkage specified, but returns user-defined type '::StructWithRef' which is incompatible with C [-Werror,-Wreturn-type-c-linkage]
   24 | ::StructWithRef cxxbridge1$199$c_return_struct_with_ref(::std::size_t const &r) noexcept {
      |                 ^
1 error generated.
...
```

# Context

This issue was originally reported in https://crrev.com/c/8342112/10#message-314ceaa401e3aca26e6dd1f4f40ce599c53aa19c

# Initial analysis

## The warning seems to indicate a real issue

Gemini tells me that on Windows, the C ABI and the C++ ABI have different rules for returning small structs by value:

* In C, any struct of size 1, 2, 4, or 8 bytes is returned directly in the RAX register. Rust's `extern "C"` ABI targets the standard C ABI, so Rust returns an 8-byte `#[repr(C)] struct StructWithRef<'a> { r: &'a usize }` in RAX.
* C++: Microsoft’s calling convention [specification says](https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170#return-values:~:text=User%2Ddefined%20types,of%20reference%20type.) that "To return a user-defined type by value in RAX [...] It can have [...] no nonstatic data members of reference type".  If a C++ struct has a reference member (e.g. `const size_t &r`), MSVC ABI classifies it as non-POD and forbids returning it in RAX, and instead uses a hidden pointer passed in RCX.

Given above, it seems that there may be a mismatch in the ABI expectations on C++-side and Rust-side - the mismatch can potentially lead to Undefined Behavior (UB).

OTOH, I haven't seen actual crashes caused by this - the arm-chair analysis above is a little bit speculative.  In particular, all the tryjobs in https://crrev.com/c/8342112/10?tab=checks are green, which seems like a potential counter-argument for the analysis above.  /cc @apasel422 (the author of https://crrev.com/c/8342112) in case they have any feedback.

## Other cases where the warning fires

This warning was encountered before - https://github.com/dtolnay/cxx/pull/1633/changes started to emit `#pragma clang diagnostic ignored "-Wreturn-type-c-linkage"` when the generated code includes a function pointer trampoline.

Gemini tells me that in case of the trampoline, it is always called within C++ - i.e. the call does not cross the FFI boundary and therefore the ABI trivially matches (because the same compiler handles the caller and the callee).

And Gemini claims that `extern "C"` was added to the trampoline's declaration solely to control name mangling for MSVC compatibility (and not for ensuring a particular ABI).

So ignoring the warning for tramplines seems okay, even if it is a bit unfortunate that the suppression covers the whole `.cc` file.

## Proposed fix

We can tweak `fn is_guaranteed_pod` in `syntax/pod.rs`, so that for `TypeQuery::Ref(_)` it returns `false` rather than `true`.  I checked that this does indeed fix the warning in the original repro outlined above.  And this seems like an okay change - in particular it doesn't change any user-visible behavior / it only changes the internal implementation details of how the `cxx`-internal thunks are declared and how they pass certain values across the FFI boundary.

anforowicz:
Because `is_guaranteed_pod` evaluated to true for structs with references, `Types::needs_indirect_abi` returned false. This caused `cxx` to generate `extern "C"` functions returning these non-POD structs directly by value (e.g. `ItemRef cxxbridge1$...()`), which triggers Clang's `-Wreturn-type-c-linkage` warning/error.

By marking `Ref(_)` as non-POD in `is_guaranteed_pod`, structs containing references correctly require indirect ABI (`void cxxbridge1$...(..., ReturnType *return$)`), resolving `-Wreturn-type-c-linkage`. Top-level `&T` return types are handled earlier in `needs_indirect_abi` and remain unaffected.

Fixes https://github.com/dtolnay/cxx/issues/1753

----------------

PTAL?

/cc @apasel422

### DIFF
--- a/syntax/pod.rs
+++ b/syntax/pod.rs
@@ -27,9 +27,9 @@ impl<'a> Types<'a> {
             | TypeQuery::SharedPtr
             | TypeQuery::WeakPtr
             | TypeQuery::CxxVector
-            | TypeQuery::Void => false,
-            TypeQuery::Ref(_)
-            | TypeQuery::Str
+            | TypeQuery::Void
+            | TypeQuery::Ref(_) => false,
+            TypeQuery::Str
             | TypeQuery::Fn
             | TypeQuery::SliceRef
             | TypeQuery::Ptr(_) => true,
--- a/tests/cpp_ui_tests.rs
+++ b/tests/cpp_ui_tests.rs
@@ -26,3 +26,45 @@ fn test_unique_ptr_of_incomplete_foward_declared_pointee() {
     let err_msg = test.compile().expect_single_error();
     assert!(err_msg.contains("definition of `::ForwardDeclaredType` is required"));
 }
+
+/// This is a regression test for returning a struct with reference members across
+/// extern "C" thunks without triggering Clang's `-Wreturn-type-c-linkage`.
+/// See also: https://github.com/dtolnay/cxx/issues/1753
+///
+/// Note that the original repro required explicitly opting into extra C/C++
+/// warnings (as already done by `.github/workflows/ci.yml`):
+///
+/// ```sh
+/// $ CXX=clang++ CXXFLAGS="-Werror -Wall -Wpedantic" cargo test --test cpp_ui_tests
+/// ```
+///
+/// This is a separate test from `tests/ffi/lib.rs` and `tests/ffi/module.rs`
+/// to avoid the risk that some other input (e.g. callback functions) will
+/// suppress the warning via `out.pragma.return_type_c_linkage = true` and
+/// `#pragma clang diagnostic ignored "-Wreturn-type-c-linkage"`.
+#[test]
+fn test_return_struct_with_ref() {
+    let test = cpp_compile::Test::new(quote! {
+        #[cxx::bridge]
+        mod ffi {
+            struct StructWithRef<'a> {
+                r: &'a usize,
+            }
+
+            unsafe extern "C++" {
+                include!("include.h");
+                fn c_return_struct_with_ref<'a>(r: &'a usize) -> StructWithRef<'a>;
+            }
+        }
+    });
+    test.write_file(
+        "include.h",
+        indoc! {"
+            #pragma once
+            #include <cstddef>
+            struct StructWithRef;
+            StructWithRef c_return_struct_with_ref(size_t const &r);
+        "},
+    );
+    test.compile().assert_success();
+}


dtolnay (approved):
Thanks!

PR #1758, opened 2026-09-11
Specialize enable_borrowed_range and enable_view for Slice<T>

### LINKED ISSUE #1757: specialize std::enable_borrowed_range for rust::Slice<T>
The c++ standard library allows you to pass ranges by value, provided `std::ranges::enable_borrowed_range<rust::Slice<T>>` is specialized to true.

### Repro:

```c++
template<typename T>
constexpr bool std::ranges::enable_borrowed_range<::rust::Slice<T>> = true;

struct MyClass
{
    rust::Vec<int> v;

    // in a real application, this function would come from rust:
    [[nodiscard]] rust::Slice<const int> get_slice() const { return rust::Slice{v}; }
};

int main()
{
    const MyClass mc{rust::Vec{1, 2, 3, 2}};

    // works:
    const auto slice = mc.get_slice();
    const std::span s1{std::ranges::find(slice, 2), slice.end()};

    // does not work without enable_borrowed_range:
    const std::span s2{std::ranges::find(mc.get_slice(), 2), mc.get_slice().end()};

    // expected result: [2, 3, 2] (from the first occurrence of 2 to the end)
}
```

### Further reading

The closest equivalent to slice is `std::span`, it specializes `enable_borrowed_range` to true: https://cppreference.com/cpp/container/span#Helper_templates

More about it on https://cppreference.com/cpp/ranges/borrowed_range.

### Solution

In my opinion, `rust::Slice` should specialize `enable_borrowed_range` to true:

```c++
template<typename T>
constexpr bool std::ranges::enable_borrowed_range<::rust::Slice<T>> = true;

// Note: this qualified syntax is not what the PR uses, as that is not supported in all compilers.
// The correct thing is to place the specialization into an enclosing namespace.
```

`rust::Slice` already references the ranges api:

```c++
// Line 281 in rust/cxx.h
#if __cplusplus >= 202002L
static_assert(std::ranges::contiguous_range<rust::Slice<const uint8_t>>);
static_assert(std::contiguous_iterator<rust::Slice<const uint8_t>::iterator>);
#endif
```

This would just be a small addition.

Note: we should probably do `ranges::enable_view`, too, as `std::span` also does that.

MixusMinimax:
Fixes #1757

### DIFF
--- a/include/cxx.h
+++ b/include/cxx.h
@@ -1147,3 +1147,17 @@ struct IsRelocatable
 
 } // namespace cxxbridge1
 } // namespace rust
+
+#if __cplusplus >= 202002L
+namespace std {
+namespace ranges {
+template <typename T>
+constexpr bool enable_borrowed_range<rust::Slice<T>> = true;
+template <typename T>
+constexpr bool enable_view<rust::Slice<T>> = true;
+} // namespace ranges
+} // namespace std
+
+static_assert(std::ranges::borrowed_range<rust::Slice<const uint8_t>>);
+static_assert(std::ranges::view<rust::Slice<const uint8_t>>);
+#endif


MixusMinimax:
I'm going to have to figure out which specific versions of the compilers support this. It is supposed to be c++20.

Edit: the problem was the qualified specialization, not the template itself.

MixusMinimax:
It's working now.

About formatting: These specializations have to be outside the namespace, so I added them to the end of the file. I could move them somewhere else, I'm not sure.

dtolnay (approved):
Thanks!

dtolnay:
Published in 1.0.201.

PR #1760, opened 2026-09-12
Do not require a writable shared header directory

### LINKED ISSUE #1752: cxx-build: Don't fail if scratch::path("cxxbridge") is not writable
I am working on a new [rust builder](https://github.com/ConradIrwin/corgi) that caches builds across many worktrees on the same machine.

To make builds more cache friendly we run build.rs in a sandbox that strictly limits file-system access (you're only allowed to write to your own $OUT_DIR). Unfortunately this currently causes cxx-bridge to fail. Target directory detection fails, and so it tries to write `cxx.h` to `scratch::path("cxxbridge")` which is disallowed.

I can think of a few ways to solve this:
* Update [make_include_dir](https://github.com/dtolnay/cxx/blob/master/bridge/build/src/lib.rs#L387-L393) to ignore failures to update the shared scratch directory (this is already the case for everything *except* `cxx.h` in [generate_bridge](https://github.com/dtolnay/cxx/blob/master/bridge/build/src/lib.rs#L419))
* Remove the fallback if target directory detection fails. For users not using cargo, there is no shared directory anymore.
* Adding an env var that allows explicit opt-in/out of the shared directory

Would you be open to any of these? 1. is probably the safest; 2. is a bit cleaner but may break other non-cargo build chains that rely on it, and 3. adds surface area that I'm not sure makes long-term sense.

gudcks0305:
Fixes #1752.

Allow `cxx-build` to finish when the optional shared header directory cannot be written, such as when a build script is restricted to writing under `OUT_DIR`.

With `DEP_CXXBRIDGE1_HEADER`, the local header link remains required and the shared link is best effort. Without it, retain the existing shared-header behavior when available, but write the embedded header directly under `OUT_DIR` if the shared write fails. A failure to create the required local header still propagates.

Four regression tests cover both header sources, a blocked shared directory, the normal shared path, and a blocked output directory. The blocked-path fixtures use a regular file in place of a directory, so they also fail deterministically when tests run as root. Both shared-directory failure cases failed before the production fix.

Validation on macOS arm64 with Rust 1.89:
- `cargo test -p cxx-build --offline`: 7 unit tests and 11 doc tests pass.
- `RUSTFLAGS='--cfg skip_ui_tests' cargo test --workspace --exclude cxx-test-suite`: passes.
- `cargo fmt --all -- --check`
- `cargo clippy -p cxx-build --all-targets --offline -- -D warnings -A unknown-lints`: passes. The exception is for pre-existing `clippy::assert_is_empty` attributes unknown to the installed toolchain.

No Windows or OS sandbox run was performed. Implemented and independently reviewed with OpenAI Codex (GPT-6 Astra, medium reasoning).

### DIFF
--- a/bridge/build/Cargo.toml
+++ b/bridge/build/Cargo.toml
@@ -29,6 +29,7 @@ syn = { version = "3", default-features = false, features = ["clone-impls", "ful
 cxx = { version = "1.0", path = "../.." }
 cxx-gen = { version = "0.7", path = "../lib" }
 pkg-config = "0.3.27"
+tempfile = "3.8"
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
--- a/bridge/build/src/lib.rs
+++ b/bridge/build/src/lib.rs
@@ -267,7 +267,8 @@ fn validate_cfg(prj: &Project) -> Result<()> {
 
 fn make_this_crate(prj: &Project) -> Result<Crate> {
     let crate_dir = make_crate_dir(prj);
-    let include_dir = make_include_dir(prj)?;
+    let header = env::var_os("DEP_CXXBRIDGE1_HEADER");
+    let include_dir = make_include_dir(prj, header.as_deref())?;
 
     let mut this_crate = Crate {
         include_prefix: Some(prj.include_prefix.clone()),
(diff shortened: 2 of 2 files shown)


dtolnay (approved):
Thanks!

dtolnay:
Published in 1.0.202.

